PR #3, opened 2022-11-07
Add a patch_constitution governance command.

plietar:
This will load the current service's constitution, split the operator-specific part from the SCITT part, and modify the latter.

Because modifying the constitution is a fairly risky procedure, there are a number of precaution measures.

### DIFF
--- a/pyscitt/pyscitt/cli/governance.py
+++ b/pyscitt/pyscitt/cli/governance.py
@@ -93,31 +93,33 @@ def propose_constitution(path: Path):
     return GovernanceAction(proposal)
 
 
-CONSTITUTION_MARKER_START = "// ----- SCITT Constitution starts here -----\n"
-CONSTITUTION_MARKER_END = "// ----- SCITT Constitution ends here -----"
+SCITT_CONSTITUTION_MARKER_START = "// ----- SCITT Constitution starts here -----\n"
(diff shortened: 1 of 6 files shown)


pyscitt/pyscitt/cli/governance.py
```diff
@@ -31,40 +63,129 @@ def execute_governance(
...
+    return GovernanceAction(path.read_bytes())
+
+
+def propose_constitution(path: Path):
+    constitution = path.read_text()
+    proposal = governance.set_constitution_proposal(constitution)
+    return GovernanceAction(proposal)
+
+
+CONSTITUTION_MARKER_START = "// ----- SCITT Constitution starts here -----\n"
```
letmaik, line 96: I think the code below is a little bit confusing semantically. It's half-way generic and half-way referring to SCITT. I think we need to decide which one it is. If it's generic, then the marker changes to something like "App Constitution Extension starts here" and there's no specific mention of scitt anywhere; maybe making the CLI command more specific is still useful like "patch_app_constitution_extension`. If the patch functionality is specific, then the variables, function names and arguments, CLI command names should probably include "scitt" more. I think I prefer the generic variant but I don't have a very strong opinion.

letmaik (changes requested):

letmaik (approved):

PR #5, opened 2022-11-09
Protect the prefix tree flush process from inconsistent KV and indexer.

plietar:
After a leader election, or a recovery, it is possible to have moved backwards from the perspective of the "continued leader", in which case the KV may contain an entry that is not known to the indexer yet.

Performing a flush in these circumstances would lead to the entry in the KV to move backwards as well, where the latest entry would have a smaller upper bound than its predecessor.

There were already asserts in batched_prefix_tree to protect against this, but these only run in debug mode, and catch the problem much too late.

We unfortunately still don't have a good way of testing these atypical interleavings and interactions between the KV and the indexer.

### DIFF
--- a/app/src/prefix_tree/frontend.h
+++ b/app/src/prefix_tree/frontend.h
@@ -147,29 +147,37 @@ namespace scitt
     {
       auto table = ctx.tx.template rw<PrefixTreeTable>(PREFIX_TREE_TABLE);
 
-      // This get() introduces an important read-dependency on any prior writes
-      // to the prefix tree table, even though we don't ever use its result.
+      auto tree = index->prepare_flush();
+
+      // It's possible for a flush to have been written to the KV, with an
+      // upper bound sequence number that has not been indexed yet. This
+      // happens for example during a leader election, where the old leader may
+      // have flushed the prefix tree, written the tree info to the KV and
+      // replicated these changes to followers. The new leader would see the
+      // up-to-date KV, but its indexer may still be lagging behind the last
+      // flush.
       //
+      // This call to get() also introduces an important read dependency.
       // Without it, the following sequence could happen, where t1 and t2 are
       // two concurrent transactions issuing a flush.
       // - The last tree to have been flushed had upper bound 4.
-      // - t1 calls start_flush(), gets upper bound 7
+      // - t1 calls prepare_flush(), gets upper bound 7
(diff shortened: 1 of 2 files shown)


plietar:
So far I've not been able to trigger this condition locally yet, but I'll try a bit harder.

app/src/prefix_tree/frontend.h
```diff
@@ -147,29 +147,37 @@ namespace scitt
     {
       auto table = ctx.tx.template rw<PrefixTreeTable>(PREFIX_TREE_TABLE);
 
-      // This get() introduces an important read-dependency on any prior writes
-      // to the prefix tree table, even though we don't ever use its result.
+      auto tree = index->prepare_flush();
+
+      // It's possible for a flush to have been written to the KV, with an
+      // upper bound sequence number that has not yet been indexed yet. This
```
letmaik, line 153:
```suggestion
      // upper bound sequence number that has not been indexed yet. This
```

app/src/prefix_tree/frontend.h
```diff
@@ -147,29 +147,37 @@ namespace scitt
...
+      // happens for example during a leader election, where the old leader may
+      // have flushed the prefix tree, written the tree info to the KV and
+      // replicated these changes to followers. The new follower would see the
```
letmaik, line 156:
```suggestion
      // replicated these changes to followers. The new leader would see the
```

letmaik (approved):

plietar:
I finally managed to trigger this case deterministically, and managed to confirm that it does happen, it does trigger assertions in debug mode and it is fixed by the PR.

The setup is a bit involved and we don't have the python test infrastructure for it at the moment. We probably can't add it to our integration tests yet, but should keep it in mind for the future.

It requires 3 nodes and a new `/app/prefix_tree/pause` HTTP endpoint that controls a boolean flag. If that flag is set, the indexing strategy's `next_requested` method returns std::nullopt, instead of the index it is interested in. This allows the indexing strategy to be paused deterministically.

Steps to reproduce:
1) Pause the indexers on the two followers
2) Submit a claim and flush the PT
3) Kill the leader process
4) Wait for a new leader to be elected
5) Submit a new claim
6) Flush the PT again
7) Resume the indexers

Without this PR, in debug mode the two nodes will crash after 7. With this PR, 6) fails gracefully with an `IndexingInProgressRetryLater` error.

Issue #8, opened 2022-11-14
Extend configuration to restrict issuers


letmaik:
Currently, the governance-controlled SCITT-specific configuration allows to set authentication options and the set of allowed signing algorithms. For the non-DID code path (not standard in SCITT at this point), the X.509 CA trust store can be configured as well, which re-uses an existing governance table from CCF.

In addition to that, it should be possible to restrict the set of allowed DID issuers. 

There are a couple of options, but I think at minimum the following should be possible (aka requirements):
- Opt-in to a set of DID methods, out of the ones that are supported. Default to allowing none.
- Allow a specific set of DIDs. Default to allowing all.

I think in the first iteration we shouldn't support regular expressions. It's not clear whether it's needed or whether this is even desirable. I seem to remember that the DID spec says somewhere that you're not supposed to deconstruct a DID and infer relationships etc.

ryazhang-microsoft:
The [PR Extend configuration to restrict issuers](https://github.com/microsoft/scitt-ccf-ledger/pull/13/files) covered requirement "Allow a specific set of DIDs. Default to allowing all".

PR #13, opened 2022-11-15
Extend configuration to restrict issuers

ryazhang-microsoft:
https://github.com/microsoft/scitt-ccf-ledger/issues/8


test/test_ccf.py
```diff
@@ -141,3 +141,39 @@ def submit(**kwargs):
...
+
+
+def test_accepted_did_issuers(tmp_path: Path):
+    def not_allowed(f):
+        with pytest.raises(ServiceError, match="InvalidInput: Unsupported did issuer in protected header"):
+            f()
+
+    with SCITTFixture(tmp_path) as fixture:
+
+        def submit(**kwargs):
```
plietar, line 153: For this test, IMO it is fine to use a single identity and signed claims throughout the test so no need for a helper function. Just define an `identity` and `claims` at the top and use that everywhere by calling `fixture.client.submit_claim` directly.

test/test_ccf.py
```diff
@@ -141,3 +141,39 @@ def submit(**kwargs):
...
+            fixture.client.submit_claim(claims)
+
+        # Start with a configuration with no accepted issuers.
+        # The service should reject anything we submit to it.
+        fixture.configure_service({"policy": {"accepted_did_issuers": []}})
+        not_allowed(lambda: submit(alg="ES256", kty="ec", ec_curve="P-256"))
+
+        # Add just one issuers to the policy. Claims signed with this
+        # issuers are accepted.
+        identity = fixture.did_web_server.create_identity(alg="ES256", kty="ec", ec_curve="P-256")
```
plietar, line 166: Using an explicit `alg`/`kty`/`ec_curve` here makes it seem like it is relevant to the current test, when actually we don't really care in this context. Removing those reduces the noise and helps highlight what actually matters.

```suggestion
        identity = fixture.did_web_server.create_identity()
```

test/test_ccf.py
```diff
@@ -141,3 +141,39 @@ def submit(**kwargs):
...
+
+        # Add just one issuers to the policy. Claims signed not with this
+        # issuers are rejected.
+        fixture.configure_service({"policy": {"accepted_did_issuers": ["else"]}})
+        not_allowed(lambda: submit(alg="ES256", kty="ec", ec_curve="P-256"))
+
+        # If no accepted_issuers are defined in the policy, any issuers
+        # is accepted.
+        fixture.configure_service({"policy": {}})
+        submit(alg="ES256", kty="ec", ec_curve="P-256")
```
plietar, line 179: There should also be a test with accepted issuers, including the one we are trying to submit.

letmaik (changes requested):

letmaik:
Code formatting job is failing, please run `./scripts/ci-checks.sh -f`

app/src/kv_types.h
```diff
@@ -96,6 +97,22 @@ namespace scitt
             std::string(JOSE_ALGORITHM_EDDSA)};
         }
       }
+      
+      bool is_accepted_issuers(std::string issuer) const
+      {
+        if (!accepted_did_issuers.has_value())
```
plietar, line 103:
```c++
bool is_accepted_issuers(std::string_view issuer) const
{
    return !accepted_did_issuers.has_value() || contains(accepted_did_issuers.value(), issuer)
}
```

app/src/kv_types.h
```diff
@@ -96,6 +97,22 @@ namespace scitt
             std::string(JOSE_ALGORITHM_EDDSA)};
         }
       }
+      
+      bool is_accepted_issuers(std::string issuer) const
```
plietar, line 101:
```suggestion
      bool is_accepted_issuers(std::string_view issuer) const
```

Using an std::string_view here avoids an unnecessary copy.

app/src/verifier.h
```diff
@@ -70,6 +70,10 @@ namespace scitt::verifier
       auto issuer = phdr.issuer;
       auto kid = phdr.kid;
       auto x5chain = phdr.x5chain;
+      if(issuer.has_value() && !configuration.policy.is_accepted_issuers(issuer.value()))
```
plietar, line 73: I would move this further down into the DID-specific code path, after we've checked that issuer has a value.

@letmaik what should we do if there is both an `x5chain` and an `issuer`? Probably we'll want to flip this logic, where we use the presence of the `issuer` field instead of `x5chain` to decide which code path we follow (since that will allow us to use `did:x509`)

test/test_ccf.py
```diff
@@ -141,3 +141,39 @@ def submit(**kwargs):
         submit(alg="ES256", kty="ec", ec_curve="P-256")
         submit(alg="ES384", kty="ec", ec_curve="P-384")
         submit(alg="PS256", kty="rsa")
+
+
+def test_accepted_did_issuers(tmp_path: Path):
```
plietar, line 146: This could probably be moved in a new `test_configuration.py` file instead. If you want, also move the `test_accepted_algorithms` into that file at the same time (it's fine if you would rather leave this for later)

plietar (changes requested):

app/src/verifier.h
letmaik: Yes, it should be flipped, let's do it in a separate PR, maybe when adding `did:x509`.

test/test_ccf.py
ryazhang-microsoft: I believe I already have one. The policy looks like {"policy": {"accepted_did_issuer": ["issuer we are trying to submit "]}}

plietar: Huh sorry my brain went wrong when I typed that. I meant "a test with *multiple* issuers".

test/test_configuration.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def test_accepted_did_issuers(tmp_path: Path):
+    def not_allowed(f):
+        with pytest.raises(ServiceError, match="InvalidInput: Unsupported did issuer in protected header"):
+            f()
+
+    with SCITTFixture(tmp_path) as fixture:
+
+        """Sign and submit the claims with a new identity"""
```
plietar, line 58: This docstring can be removed.

app/src/kv_types.h
```diff
@@ -73,11 +73,14 @@ namespace scitt
        * List of accepted COSE signature algorithms when verifying signatures.
        * The names are case sensitive.
        *
+       * List of accepted DID issuer when verifying signatures.
+       * The names are case sensitive.
+       *
```
plietar, line 78: You should make a separate comment block for the new field.

plietar (changes requested):
Overall LGTM, just some minor nits left.

You should also update the constitution to apply some validation to the input:
https://github.com/microsoft/scitt-ccf-ledger/blob/31301a7de8fd662f6c05cb6225208c9b229153f8/app/constitution/scitt.js#L15-L20

app/src/kv_types.h
```diff
@@ -72,12 +72,17 @@ namespace scitt
...
        * Rather than the COSE integer algorithm IDs, we use the equivalent
        * human-friendly JOSE names.
        */
       std::optional<std::vector<std::string>> accepted_algorithms;
 
+      /**
+       * List of accepted DID issuer when verifying signatures.
+       * The names are case sensitive.
+       */
+      std::optional<std::vector<std::string>> accepted_did_issuer;
```
letmaik, line 84: This should be `accepted_did_issuers`, why singular?

Issue #11, opened 2022-11-14
Check C++ code with scan-build


letmaik:
https://clang-analyzer.llvm.org/scan-build.html

See also the CI pipeline scripts in https://github.com/microsoft/CCF on how scan-build is integrated there.

letmaik:
Done in #43.

PR #24, opened 2022-11-23
scan-build integration

ryazhang-microsoft:
https://github.com/microsoft/scitt-ccf-ledger/issues/11


.pipelines/azure-pipelines.yml
```diff
@@ -111,7 +111,14 @@ stages:
 
           - script: ./scripts/ci-checks.sh
             displayName: "CI checks"
-  
+
+      - job: scanbuild
+        displayName: Static Analytics
```
letmaik, line 116:
```suggestion
        displayName: Static Analysis 
```

scripts/scan-build.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
letmaik, line 1: Wouldn't something like the following work?
```
CLANG_VERSION=10
ENCLAVE_TYPE=virtual CMAKE_BUILD_TYPE=Debug scan-build-$CLANG_VERSION --use-cc=clang-$CLANG_VERSION --use-c++=clang++-$CLANG_VERSION ./build.sh
```

letmaik (approved):

PR #27, opened 2022-11-24
Add an isolated_test marker for functional tests.

plietar:
If a test class is annotated with the marker, a separate cchost process is started for this test class. This provides isolation between tests, which is useful either if the test is deemed risky and can leave the service in an irrecoverably bad state (eg. by modifying the constitution), or if the test depends on having an empty ledger at start.

Note that when using an external service, it is impossible to provide isolation and even tests marked with isolated_test will use the shared instance. The marker can be combined with the needs_cchost marker to skip the test in these cases where isolation cannot be guaranteed.

### DIFF
--- a/test/infra/fixtures.py
+++ b/test/infra/fixtures.py
@@ -60,7 +60,7 @@ class ManagedCCHostFixtures:
         Start a managed SCITT service, using cchost.
 
         This fixture returns a function, which creates a new service everytime
-        it is called. Test should bot use this fixture directly. Instead the
+        it is called. Tests should not use this fixture directly. Instead the
         cchost (or client) fixture should be used, which provides access to an
         already running instance.
         """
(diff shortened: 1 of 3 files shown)


test/infra/fixtures.py
```diff
@@ -31,15 +32,38 @@ def __init__(self, binary, enclave_type, enclave_file, constitution):
...
     @pytest.fixture(scope="session")
-    def cchost(self, tmp_path_factory):
+    def start_cchost(self, tmp_path_factory):
         """
         Start a managed SCITT service, using cchost.
 
-        The service will keep running for the entire test session, and therefore
-        shared among tests.
+        This fixture returns a function, which creates a new service everytime
+        it is called. Test should bot use this fixture directly. Instead the
```
letmaik, line 63:
```suggestion
        it is called. Tests should not use this fixture directly. Instead the
```

letmaik (approved):

PR #35, opened 2022-12-07
Add iss and kid fields to receipts.

### LINKED ISSUE #25: Add `iss` and `kid` field to receipts
We're headed towards using a DID for the server's identity. This will require a few parts, but the first step would be to add `iss` and `kid` fields to the receipt's protected headers.

The `iss` value would be copied from an optional field in the configuration file. If the field is missing, we'll omit the `iss` and `kid` altogether. The `kid` should represent the current service's certificate/public key and will change on disaster recovery. For now, I think we can use the same value we use for `service_identity`, ie. the hash of the DER certificate, and revisit this in the future.

Even though it will be redundant with the `kid`, I think we shouldn't remove the `service_identity` field just yet, so as to not break tooling that uses it while we are still transitioning to DIDs.

plietar:
We're headed towards using a DID for the server's identity. This will require a few parts, but the first step is to add iss and kid fields to the receipt's protected headers.

The iss value is be copied from an optional field in the configuration file. If the field is missing, we omit the iss and kid altogether. The kid should represent the current service's certificate/public key and will change on disaster recovery. For now, we use the same value we use for service_identity, ie. the hash of the DER certificate, but may revisit this in the future (eg. a hash of just the public key).

The old service_identity field does not get removed just yet, so as to not break tooling that uses it while we are still transitioning to DIDs. In particular, pyscitt and our functional tests still depend on it.

Fixes #25

### DIFF
--- a/app/src/kv_types.h
+++ b/app/src/kv_types.h
@@ -127,6 +127,10 @@ namespace scitt
 
     Policy policy = {};
     Authentication authentication = {};
+
+    // The long-term stable identity of this service, as a DID.
+    // If set, it will be used to populated the issuer field of receipts
+    std::optional<std::string> service_identity;
   };
 
   DECLARE_JSON_TYPE_WITH_OPTIONAL_FIELDS(Configuration::Policy);
@@ -146,7 +150,8 @@ namespace scitt
 
   DECLARE_JSON_TYPE_WITH_OPTIONAL_FIELDS(Configuration);
   DECLARE_JSON_REQUIRED_FIELDS(Configuration);
-  DECLARE_JSON_OPTIONAL_FIELDS(Configuration, policy, authentication);
+  DECLARE_JSON_OPTIONAL_FIELDS(
+    Configuration, policy, authentication, service_identity);
 
   // Tables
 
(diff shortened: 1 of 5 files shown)


app/src/kv_types.h
```diff
@@ -127,6 +127,10 @@ namespace scitt
 
     Policy policy = {};
     Authentication authentication = {};
+
+    // The long-term stable identity of this service, as a DID.
+    // If set, it will be used to populated the issuer field of receipts
```
letmaik, line 132:
```suggestion
    // If set, it will be used to populate the issuer field of receipts
```

test/test_configuration.py
```diff
@@ -99,3 +100,34 @@ def test_multiple_accepted_issuers(
             {"policy": {"accepted_did_issuers": [identity.issuer, "else"]}}
         )
         client.submit_claim(claims)
+
+
+class TestServiceIdentifier:
+    @pytest.fixture
+    def claims(self, did_web):
```
letmaik, line 107: Should be singular `claim`.

letmaik (approved):

PR #40, opened 2022-12-14
Add integration test for non-canonically encoded protected headers.

plietar:
The COSE spec does not impose any encoding requirements for the contents of the protected headers, but instead requires that these are wrapped in a bstr and never re-encoded by any party processing the message.

This adds some sanity checks, mostly focusing on these points:
- The ledger accepts claims with a non-canonical encoding
- When embedding a receipt, the ledger does not affect the encoding of the claim.
- pyscitt can verify the receipts for these claims.

The last point is actually broken for now, due to a bug in the current pycose release. This has been fixed upstream, but hasn't been released yet. The relevant test is marked as XFAIL, until pycose is updated.

### DIFF
--- a/test/test_encoding.py
+++ b/test/test_encoding.py
@@ -12,7 +12,7 @@ from pyscitt import crypto
 
 class TestNonCanonicalEncoding:
     @pytest.fixture
-    def claims(self, did_web):
+    def claim(self, did_web):
         """Create a signed claim, which protected headers encoded non-canonically."""
 
         identity = did_web.create_identity()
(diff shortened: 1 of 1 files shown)


test/test_encoding.py
```diff
@@ -0,0 +1,76 @@
...
+from cryptography.exceptions import InvalidSignature
+from pycose import algorithms, headers
+from pycose.messages import Sign1Message
+
+from pyscitt import crypto
+
+
+class TestNonCanonicalEncoding:
+    @pytest.fixture
+    def claims(self, did_web):
```
letmaik, line 15: Should be `claim` I think.

test/test_encoding.py
```diff
@@ -0,0 +1,76 @@
...
+    def test_embed_receipt(self, client, trust_store, claims):
+        """When embedding a receipt in a claim, the ledger should not affect the original encoding."""
+        tx = client.submit_claim(claims, decode=False).tx
+        embedded = client.get_claim(tx, embed_receipt=True)
+
+        original_pieces = cbor2.loads(claims).value
+        updated_pieces = cbor2.loads(embedded).value
+
+        # Any part of the message that is cryptographically bound needs to be preserved.
+        # These are respestively, the protected header, the payload and the signature.
```
letmaik, line 73:
```suggestion
        # These are respectively, the protected header, the payload and the signature.
```

test/test_encoding.py
```diff
@@ -0,0 +1,76 @@
...
+        signature = algorithms.Es256.sign(key, tbs)
+        message = [protected, dict(), payload, signature]
+        return cbor2.dumps(cbor2.CBORTag(Sign1Message.cbor_tag, message))
+
+    def test_submit_claim(self, client, trust_store, claims):
+        """The ledger should accept claims even if not canonically encoded."""
+        client.submit_claim(claims, decode=False)
+
+    @pytest.mark.xfail(
+        reason="pytest does not preserve the original encoding", raises=InvalidSignature
```
letmaik, line 56:
```suggestion
        reason="pycose does not preserve the original encoding", raises=InvalidSignature
```

letmaik (approved):

PR #41, opened 2022-12-14
Add integration tests for the endpoints that deal with old claims.

plietar:
The ledger allows old claims and receipts to be enumerated and retrieved. At one point (before the open source release) this code path was actually broken and subsequently fixed, but I had never added the integration tests.


letmaik (approved):

letmaik:
CI is failing for `test_enumerate_claims`. The enumeration seems to contain a lot more claims from the start (of other tests in this case), but the end of the list is cut off in the log output, so I'm not sure if that's correct either.

plietar:
Aha, silly me had only tried running just that test, and never the whole suite.
Seems like the start parameter isn't working correctly, I'll look into it.

plietar:
The reason turned out to be that the endpoint expects integer sequence numbers, rather than a transaction ID (which is two number separated with a dot), but was silently ignoring ill-formed values and treating them as if they were missing.

I've changed the C++ code to be a bit stricter, and return an error in these cases.

app/src/main.cpp
```diff
@@ -348,121 +388,120 @@ namespace scitt
...
-        if (to_seqno < from_seqno)
-        {
-          throw BadRequestError(
-            errors::InvalidInput,
-            fmt::format(
-              "Invalid range: Starts at {} but ends at {}",
-              from_seqno,
-              to_seqno));
-        }
+          if (to_seqno < from_seqno)
```
plietar, line 420: clang-format works in mysterious ways, and somehow decided it was time to re-ident everything in this lambda :(

PR #42, opened 2022-12-14
Add /version endpoint

### LINKED ISSUE #36: Add an endpoint which returns the version and commit hash
When debugging issues with a running service, it can be useful to know which version is running. We should embed this at build time and expose it over the HTTP interface. This already exists in CCF as the `/node/version` endpoint, which returns the version of the framework.

We should make it clear what the behaviour is when multiple versions are running. We may want both a `/app/node_version` and a `/app/leader_version` endpoint, where the former never gets forwarded and the latter always does.

This is of course a self-reported value, used for debugging only. It does not replace the existing code identity managed by CCF.

letmaik:
Closes #36.


ryazhang-microsoft (approved):

app/cmake/version.cmake
```diff
@@ -0,0 +1,25 @@
...
+unset(SCITT_VERSION)
+
+if (SCITT_VERSION_OVERRIDE)
+  set(SCITT_VERSION "${SCITT_VERSION_OVERRIDE}")
+else()
+  # Deduce project version from git environment
+  find_package(Git)
+
+  execute_process(
+    COMMAND "bash" "-c" "${GIT_EXECUTABLE} describe --tags --match=\"*.*.*\""
```
plietar, line 13: Is `bash -c` necessary? Wouldn't just `COMMAND "${GIT_EXECUTABLE}" "describe" "--tags" "--match=*.*.*"` work?

plietar (approved):

PR #43, opened 2022-12-14
scan-build integration #24

ryazhang-microsoft:
https://github.com/microsoft/scitt-ccf-ledger/issues/11


letmaik:
The CI shows no bugs found by scan-build. Did you verify manually that bugs are actually reported? There also seems to be a distinction between source and header files. Might be worth adding a bug that scan-build detects in some header that's only included by another header file, like `src/did/web/syntax.h`, just to make sure this is actually working as it should.

letmaik (approved):

Issue #31, opened 2022-11-30
Sign governance proposals via Key Vault


letmaik:
https://microsoft.github.io/CCF/main/governance/hsm_keys.html#http-request-signature

The idea is to extend the Python CLI to talk to Key Vault in addition to supporting local keys.

PR #54, opened 2023-01-03
Sign governance proposals with AKV

ryazhang-microsoft:
Closes https://github.com/microsoft/scitt-ccf-ledger/issues/31


pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,101 @@
...
+
+        styled_cert = "-----BEGIN CERTIFICATE-----\n"
+        for index in range(0, len(decoded), 64):
+            if index + 64 > len(decoded):
+                styled_cert += decoded[index:]
+            else:
+                styled_cert += decoded[index : index + 64]
+            styled_cert += "\n"
+        styled_cert += "-----END CERTIFICATE-----"
+        return styled_cert
```
letmaik, line 51: Let's keep this simple:

```suggestion
        cert_pem = f"-----BEGIN CERTIFICATE-----\n{decoded}\n-----END CERTIFICATE-----"
        return cert_pem
```

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def sign_with_idetity(self, digest: bytes):
```
letmaik, line 53: Typo `identity`

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,101 @@
...
+        # The signatures returned by AKV are returned as a JWS signature and encoded in
+        # base64url format and are not directly compatible with the signatures supported by CCF.
+        sign_result = None
+        signature_algorithm = SignatureAlgorithm.es384
+        # See https://github.com/microsoft/CCF/blob/master/doc/members/jws_to_der.py
+        # for original conversion code.
+        # digest_to_sign = hashlib.sha384(digest.encode()).digest()
+
+        digest_to_sign = {
+            256: hashlib.sha256(digest).digest(),
```
letmaik, line 79: We should just use the default alg that matches the curve of the key. See also `default_algorithm_for_key` in the crypto module which takes a public or private key as input. In this case, it would be the public key from the cert.

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,101 @@
...
+    """Internal helper class for decoding AKV signature for CCF."""
+    componentType = NamedTypes(
+        NamedType("r", Integer()),
+        NamedType("s", Integer()),
+    )
+
+class KeyVaultSignClient():
+    """MemberIdentity implementation that uses Azure Key Vault."""
+
+    def __init__(self, akv_sign_configuration: str):
```
letmaik, line 31: This should take the decoded JSON as dict.

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,101 @@
...
+        akv_sign_configuration_dict = json.loads(akv_sign_configuration)
+        self._vault_name = akv_sign_configuration_dict['keyVaultName']
+        self._vault_url = f"https://{self._vault_name}.vault.azure.net"
+        self._identity_certificate_name = akv_sign_configuration_dict['cetificateName']
```
letmaik, line 35: Typo `certificate` in this line and below.

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,101 @@
...
+        self._identity_certificate_version = akv_sign_configuration_dict['cetificateVersion']
+        self._identity_private_curve_key_size = akv_sign_configuration_dict['privateCurveKeySize']
```
letmaik, line 37: Not needed, see comment below.

pyscitt/pyscitt/client.py
```diff
@@ -27,11 +29,14 @@
 class HttpSig(httpx.Auth):
     requires_request_body = True
 
-    def __init__(self, key_id: str, pem_private_key: str):
-        self.key_id = key_id
-        self.private_key = load_pem_private_key(
-            pem_private_key.encode("ascii"), password=None, backend=default_backend()
-        )
+    def __init__(self, key_id: str, pem_private_key: str, akv_configuration:str):
```
letmaik, line 32: There should be no AKV-related code in this class. Rather, `key_id` and `pem_private_key` should be replaced by some signer object with a uniform interface for which we have two implementations, local key and AKV.

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,101 @@
...
+        cert = cert_client.get_certificate_version(
+            certificate_name=self._identity_certificate_name,
+            version=self._identity_certificate_version
+            )
+
+        crypto_client = CryptographyClient(key, credential=credential)
+
+        # The signatures returned by AKV are returned as a JWS signature and encoded in
+        # base64url format and are not directly compatible with the signatures supported by CCF.
+        sign_result = None
```
letmaik, line 72: No need to define `sign_result` here.

pyscitt/pyscitt/client.py
```diff
@@ -78,7 +94,7 @@ class BaseClient:
 
     url: str
     auth_token: Optional[str]
-    member_auth: Optional[Tuple[str, str]]
+    member_auth: Optional[Tuple[str, str, str]]
```
plietar, line 97: I think this would be cleaner by defining an abstract class, with two implementations (KeyVaultSignClient and a local credentials one).

For example:
```
class MemberAuthenticationMethod(ABC):
    def sign_proposal(self, data: bytes) -> bytes: ...
```

and then in BaseClient, the field would be `member_auth: MemberAuthenticationMethod`.

The `client_arguments.py` file would create the right object depending on the arguments.

pyscitt/pyscitt/client.py
```diff
@@ -45,12 +55,18 @@ def auth_flow(self, request):
...
-            self.private_key.curve.key_size
-        ]
-        signature = self.private_key.sign(
-            signature_algorithm=ec.ECDSA(algorithm=digest_algo), data=string_to_sign
-        )
+
+        if hasattr(self, "akv_configuration"):
+            key_vault_sign_client = KeyVaultSignClient(self.akv_configuration)
+            signature, cert = key_vault_sign_client.sign_with_identity(string_to_sign)
+            self.key_id = crypto.get_cert_fingerprint(cert)
```
plietar, line 62: There shouldn't be a need to mutate self after the HttpSig was created.

pyscitt/pyscitt/cli/client_arguments.py
```diff
@@ -67,6 +67,12 @@ def add_client_arguments(
             type=Path,
             help="Path to CCF member certificate file",
         )
+        parser.add_argument(
+            "--akv-configuration",
+            default=None,
```
plietar, line 72: `default=None` is already implicit IIRC, so no need to add it.

pyscitt/pyscitt/cli/client_arguments.py
```diff
@@ -88,6 +94,10 @@ def create_client(args: argparse.Namespace):
         key = args.member_key.read_text()
         kwargs["member_auth"] = (cert, key)
 
+    if "akv_configuration" in args and args.akv_configuration:
+        akv_configuration = args.akv_configuration.read_text()
+        kwargs["member_auth"] = (None, None, akv_configuration)
+
```
plietar, line 100: As it is now, even if an AKV configuration is specified it will try to read the cert and key at the default paths, and fail if the files don't exist.

I think we should check AKV first, and only if it is not specified try to read the cert/key.

pyscitt/pyscitt/client.py
```diff
@@ -27,11 +29,19 @@
 class HttpSig(httpx.Auth):
     requires_request_body = True
 
-    def __init__(self, key_id: str, pem_private_key: str):
-        self.key_id = key_id
-        self.private_key = load_pem_private_key(
-            pem_private_key.encode("ascii"), password=None, backend=default_backend()
-        )
+    def __init__(self, key_id: str, pem_private_key: str, akv_configuration: str):
```
darracott, line 32: Might be worth making the `key_id` an `Optional[str]`, see https://mypy.readthedocs.io/en/stable/kinds_of_types.html#optional-types-and-the-none-type

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,105 @@
...
+        # for original conversion code.
+        jws_raw = sign_result.signature
+        jws_raw_len = len(jws_raw)
+
+        der_signature = _DERSignature()
+        der_signature["r"] = int.from_bytes(
+            jws_raw[: int(jws_raw_len / 2)], byteorder="big"
+        )
+        der_signature["s"] = int.from_bytes(
+            jws_raw[-int(jws_raw_len / 2) :], byteorder="big"
```
plietar, line 96: I guess the `-` works, but I find it somewhat confusing. Would `jws_raw[jws_raw_len // 2: ]` should work just the same (assuming jws_len has an even number of bytes)

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+        return encode(der_signature)
```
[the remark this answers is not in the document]
plietar: I've actually had to implement that exact logic in my own PR: https://github.com/microsoft/scitt-ccf-ledger/pull/61/files#diff-a4283a0a03024d48c620ec7878ac696fe3e6313d154b8fce00bbe7bc3e6d0a7cR180-R209

Once either PR gets merged we can refactor this a bit and avoid the duplication.

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def get_key_id(self) -> bytes:
+        return self.key_id
+
+    def sign(self, data: bytes) -> bytes:
+        signature = self.sign_with_identity(data)
+        return signature
```
plietar, line 105: Is this method necessary? `sign_with_identity` can be renamed to `sign`.

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+        self.key_id = crypto.get_cert_fingerprint(self.cert)
+
+    @staticmethod
+    def _decode_certificate(cert: KeyVaultCertificate) -> str:
+        decoded = base64.b64encode(cert.cer).decode()  # type: ignore
+        cert_pem = f"-----BEGIN CERTIFICATE-----\n{decoded}\n-----END CERTIFICATE-----"
+        return cert_pem
+
+    def sign_with_identity(self, digest: bytes):
```
plietar, line 57:
```suggestion
    def sign(self, data: bytes):
```

Using `digest` here is misleading. The argument is the actual request, not a hash of it.

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,105 @@
...
+        # credential = DefaultAzureCredential()
```
plietar, line 58: Line can be removed

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,105 @@
...
+        # See https://github.com/microsoft/CCF/blob/master/doc/members/jws_to_der.py
+        # for original conversion code.
+        # digest_to_sign = hashlib.sha384(digest.encode()).digest()
+        signature_algorithm = SignatureAlgorithm.es384
+        cert = load_pem_x509_certificate(self.cert.encode("ascii"), default_backend())
+
+        digest_to_sign = {
+            "sha256": hashlib.sha256(digest).digest(),
+            "sha384": hashlib.sha384(digest).digest(),
+        }[cert.signature_hash_algorithm.name]
```
plietar, line 79: This is a bit odd in that it will compute both the sha256 and sha384, and then select one result or the other. Instead it should select one algorithm, and compute that one only.

Alternatively, this should work:
```suggestion
        digest_to_sign = hashlib.new(cert.signature_hash_algorithm.name, data).digest()
```

pyscitt/pyscitt/client.py
```diff
@@ -15,22 +16,31 @@
...
-from . import crypto
 from .governance import GovernanceClient
 from .prefix_tree import PrefixTreeClient
 from .receipt import Receipt
 
 CCF_TX_ID_HEADER = "x-ms-ccf-transaction-id"
 
+
+class MemberAuthenticationMethod(ABC):
+    @property
```
plietar, line 27: A `@property` means this method should look and feel just like a field from the outside (read-only in this case). It should be called just `key_id`, without the `get_` prefix, and should be accessed as if it were a field, not a method, ie. `member_auth_client.key_id` instead of `member_auth_client.get_key_id()` (note the lack of parenthesis).

The subclasses should either have methods that are also marked as `@property` themselves, or they should just be plain fields.

eg.
```python
class MemberAuthenticationMethod(ABC):
    @property
    @abstractmethod
    def key_id(self) -> bytes: pass

# either
class LocalKeySignClient(MemberAuthenticationMethod):
    key_id: bytes

    def __init__(self, [...]):
        self.key_id = [...]

# or
class LocalKeySignClient(MemberAuthenticationMethod):
    _key_id: bytes

    def __init__(self, [...]):
        self._key_id = [...]
 
    @property
    def key_id(self) -> bytes:
        return self._key_id
```


(It would also be fine to make this a normal method and drop the `@property` altogether.)

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,80 @@
...
+from azure.keyvault.keys import KeyClient
+from azure.keyvault.keys.crypto import CryptographyClient, SignatureAlgorithm
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.asymmetric.utils import encode_dss_signature
+from cryptography.x509 import load_pem_x509_certificate
+
+from . import crypto
+
+
+class KeyVaultSignClient:
```
plietar, line 18:
```suggestion
class KeyVaultSignClient(MemberAuthenticationMethod):
```

You should inherit from the abstract class.

pyscitt/pyscitt/client.py
```diff
@@ -15,22 +16,31 @@
...
     requires_request_body = True
 
-    def __init__(self, key_id: str, pem_private_key: str):
-        self.key_id = key_id
-        self.private_key = load_pem_private_key(
-            pem_private_key.encode("ascii"), password=None
-        )
+    def __init__(self, member_auth_client: MemberAuthenticationMethod):
+        self.member_auth_client = member_auth_client
+        self.key_id = member_auth_client.get_key_id()
```
plietar, line 43: I don't think it is useful to cache the key id, especially since neither auth client make it expensive to retrieve it. You can just look it up as `self.member_auth_client.key_id` each time `auth_flow` is called. Not a big deal though.

pyscitt/pyscitt/client.py
```diff
@@ -15,22 +16,31 @@
...
+        pass
+
+    @abstractmethod
+    def sign(self, data: bytes) -> bytes:
+        pass
+
+
 # copied from CCF/tests/infra/clients.py
 class HttpSig(httpx.Auth):
     requires_request_body = True
```
plietar, line 39:
```suggestion
class HttpSig(httpx.Auth):
    member_auth_client: MemberAuthenticationMethod

    requires_request_body = True
```

It's a good idea to have the field type.

pyscitt/pyscitt/client.py
```diff
@@ -88,7 +94,7 @@ def __init__(
         url: str,
         *,
         auth_token: Optional[str] = None,
-        member_auth: Optional[Tuple[str, str]] = None,
+        member_auth: MemberAuthenticationMethod = None,
```
plietar, line 97:
```suggestion
        member_auth: Optional[MemberAuthenticationMethod] = None,
```

pyscitt/pyscitt/client.py
```diff
@@ -77,7 +83,7 @@ class BaseClient:
 
     url: str
     auth_token: Optional[str]
-    member_auth: Optional[Tuple[str, str]]
+    member_auth: MemberAuthenticationMethod
```
plietar, line 86:
```suggestion
    member_auth: Optional[MemberAuthenticationMethod]
```

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,80 @@
...
+        self.cert = self._decode_certificate(
+            cert_client.get_certificate_version(
+                certificate_name=self._identity_certificate_name,
+                version=self._identity_certificate_version,
+            )
+        )
+        self.key_id = crypto.get_cert_fingerprint(self.cert)
+
+    @staticmethod
+    def _decode_certificate(cert: KeyVaultCertificate) -> str:
```
plietar, line 41:
```suggestion
    def _encode_certificate(cert: KeyVaultCertificate) -> str:
```

Kind of pendantic, but IMO this is *encoding* rather than decoding. It's taking the raw cert bytes and making it into PEM format.

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,80 @@
...
+
+        crypto_client = CryptographyClient(key, credential=self.credential)
+
+        # The signatures returned by AKV are returned as a JWS signature and encoded in
+        # base64url format and are not directly compatible with the signatures supported by CCF.
+        # See https://github.com/microsoft/CCF/blob/master/doc/members/jws_to_der.py
+        # for original conversion code.
+        # digest_to_sign = hashlib.sha384(digest.encode()).digest()
+        signature_algorithm = SignatureAlgorithm.es384
+        cert = load_pem_x509_certificate(self.cert.encode("ascii"), default_backend())
```
plietar, line 62:
```suggestion
        cert = load_pem_x509_certificate(self.cert.encode("ascii"))
```

See https://github.com/microsoft/scitt-ccf-ledger/pull/62

pyscitt/pyscitt/local_key_sign_client.py
```diff
@@ -0,0 +1,36 @@
...
+class LocalKeySignClient(MemberAuthenticationMethod):
+    def __init__(self, cert: str, key: str) -> None:
+        self.cert = cert
+        self.key = key
+        self.key_id = crypto.get_cert_fingerprint(cert)
+        self.private_key = load_pem_private_key(
+            key.encode("ascii"),
+            password=None,
+            backend=default_backend(),
+        )
```
plietar, line 24:
```suggestion
        self.private_key = load_pem_private_key(
            key.encode("ascii"),
            password=None,
        )
```

See https://github.com/microsoft/scitt-ccf-ledger/pull/62

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+
+        # return sign_result, key.id
+        # See https://github.com/microsoft/CCF/blob/master/doc/members/jws_to_der.py
+        # for original conversion code.
+        jws_raw = sign_result.signature
+        jws_raw_len = len(jws_raw)
+        r = int.from_bytes(jws_raw[: jws_raw_len // 2], byteorder="big")
+        s = int.from_bytes(jws_raw[jws_raw_len // 2 :], byteorder="big")
+        return encode_dss_signature(r, s)
```
plietar, line 77:
```suggestion
        return crypto.convert_p1363_signature_to_dss(sign_result.signature)
```

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,80 @@
...
+
+        # The signatures returned by AKV are returned as a JWS signature and encoded in
+        # base64url format and are not directly compatible with the signatures supported by CCF.
+        # See https://github.com/microsoft/CCF/blob/master/doc/members/jws_to_der.py
+        # for original conversion code.
+        # digest_to_sign = hashlib.sha384(digest.encode()).digest()
+        signature_algorithm = SignatureAlgorithm.es384
+        cert = load_pem_x509_certificate(self.cert.encode("ascii"), default_backend())
+
+        digest_to_sign = hashlib.new(cert.signature_hash_algorithm.name, data).digest()
```
plietar, line 64: I'm not sure this is right actually. `cert.signature_hash_algorithm` is the hash algorithm that was used to sign the certificate when it was issued. This is not necessarily the same hash algorithm we want to be using to hash payloads.

You already hardcode ES384 above, so hardcoding sha384 using `hashlib.sha384(data).digest()` is probably fine here too. Sorry I should have caught that in the previous review pass.

If you don't want to hardcode es384/sha384, then the size of the certificate's key is a better way to determine both the signature and hash algorithms. `cert.public_key().curve.key_size` should give you the key size, which can then be used to select the algorithms, similar to how it is done in `LocalKeySignClient`.

https://cryptography.io/en/latest/x509/reference/#cryptography.x509.Certificate.signature_hash_algorithm

pyscitt/pyscitt/local_key_sign_client.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self, cert: str, key: str) -> None:
+        self.cert = cert
+        self.key = key
+        self.private_key = load_pem_private_key(
+            key.encode("ascii"),
+            password=None,
+        )
+
+    def sign(self, data: bytes) -> bytes:
+        print(self.private_key)
```
plietar, line 22: This can be removed

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,63 @@
...
+        self.cert = self._encode_certificate(
+            cert_client.get_certificate_version(
+                certificate_name=self._identity_certificate_name,
+                version=self._identity_certificate_version,
+            )
+        )
+
+    @staticmethod
+    def _encode_certificate(cert: KeyVaultCertificate) -> str:
+        decoded = base64.b64encode(cert.cer).decode()  # type: ignore
```
plietar, line 44: I'm curious why this needs a `type: ignore`?

pyscitt/pyscitt/client.py
```diff
@@ -22,15 +23,21 @@
 
 CCF_TX_ID_HEADER = "x-ms-ccf-transaction-id"
 
+
+class MemberAuthenticationMethod(ABC):
```
plietar, line 27: This should have a `cert: str` property, since it gets used by HttpSig.

pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -0,0 +1,65 @@
...
+        key = key_client.get_key(
+            name=self._identity_certificate_name,
+            version=self._identity_certificate_version,
+        )
+        crypto_client = CryptographyClient(key, credential=self.credential)
+        cert = load_pem_x509_certificate(self.cert.encode("ascii"))
+        assert isinstance(cert.public_key, (EllipticCurvePublicKey))
+        key_size = cert.public_key().curve.key_size
+        signature_algorithm = "ES" + str(key_size)
+        hash_algorithm = "sha" + str(key_size)
```
plietar, line 60: Confusingly, a 521 bit curve (ie. P521) uses ES512 (ie. sha512). Note that the 1 and 2 are flipped.

Using a dictionary lookup like in the local client would work:
```suggestion
ALGORITHMS = { 256: ("ES256", "sha256"}, 384: ("ES384", "sha384"), 521: ("ES512", "sha384") }
signature_algorithm, hash_algorithm = ALGORITHMS[key_size]
```

plietar (approved):
Just a minor point left. Otherwise LGTM

Issue #37, opened 2022-12-12
Revive load testing and run in daily CI


letmaik:
https://github.com/microsoft/scitt-ccf-ledger/tree/main/test/load_test hasn't received much attention in a while and may be broken by now as it's not tested in CI.

The scripts should be checked and a new daily CI (like in CCF) added that runs the load test.

letmaik:
I think a good option for this is to use locust as a library (https://docs.locust.io/en/stable/use-as-lib.html). This would allow us to write it as a regular pytest and re-use all the fixtures.

PR #60, opened 2023-01-05
Add mypy checks to CI.

### LINKED ISSUE #10: Check types in Python code with mypy
This should cover both `pyscitt/` and `test/`.

plietar:
Fixes #10


test/load_test/locustfile.py
```diff
@@ -1,139 +0,0 @@
-# Copyright (c) Microsoft Corporation.
```
letmaik, line 1: See https://github.com/microsoft/scitt-ccf-ledger/issues/37. Can we keep those files and disable mypy validation via some annotation until we fix them?

letmaik (approved):

PR #63, opened 2023-01-06
Unset SGX_AESM_ADDR env variable when starting cchost process

darracott:
Open Enclave fails to start when SGX_AESM_ADDR is set. This PR ensures it is not set when starting SCITT. 

SGX_AESM_ADDR not required for CCF SCITT and is described here https://learn.microsoft.com/en-us/azure/confidential-computing/confidential-nodes-aks-addon#overview

### DIFF
--- a/test/infra/cchost.py
+++ b/test/infra/cchost.py
@@ -192,6 +192,10 @@ class CCHost:
                 raise
 
     async def _start_process(self):
+        # Ensure SGX_AESM_ADDR is not set when starting cchost.
+        cchost_env = os.environ.copy()
+        cchost_env.pop("SGX_AESM_ADDR", None)
+        
         LOG.debug("Starting cchost process...")
         process = await asyncio.create_subprocess_exec(
             self.binary,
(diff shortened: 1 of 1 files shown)


test/infra/cchost.py
```diff
@@ -192,6 +192,10 @@ async def _start_service(self):
                 raise
 
     async def _start_process(self):
+        # Ensure SGX_AESM_ADDR is not set when starting cchost.
+        cchost_env = os.environ.copy()
+        cchost_env.pop("SGX_AESM_ADDR", None)
```
plietar, line 197:
```suggestion
        cchost_env.pop("SGX_AESM_ADDR")
```

plietar (approved):

test/infra/cchost.py
darracott: Without a default set (here `None`) then `pop` will fail if SGX_AESM_ADDR is already unset (or was never set)

plietar: Oh my bad, I somehow assumed `None` was already the default value (I think I mixed it up with the `get` method).

PR #67, opened 2023-01-10
Use relative DID URL as kid

### LINKED ISSUE #22: Make `kid` a relative DID URL
The DID/VC community generally used absolute DID URLs as `kid` in VC-JWTs but they are moving to relative to avoid faulty implementations that forget to match that the issuer is a prefix of `kid`. See https://github.com/decentralized-identity/jwt-vc-presentation-profile/blob/main/spec/spec.md#kid-jwt-header for example.

In our code, `kid` is the fragment value, so not an actual relative DID URL. This should be changed to to start with `#` to be consistent and avoid interop issues later on. This aspect is currently underspecified in the SCITT specs.

Note that DID resolution options, if any, are not meant to go into `kid`, so this is a restricted form of relative DID URL only allowing a fragment.

letmaik:
Fixes #22.

Note that this is a breaking change when did:web is used as issuer in submitted claims. Clients must now prefix the key id with `#` making it a relative DID URL.


test/test_ccf.py
```diff
@@ -113,3 +113,43 @@ def test_default_did_port(client, trust_store, tmp_path):
...
+
+    # Check that the resolved DID document contains the expected assertion
+    # method id.
+    did_doc = client.get_did_document(identity.issuer)
+    assert did_doc["assertionMethod"][0]["id"] == f"{identity.issuer}{kid}"
+
+
+def test_invalid_kid(client, did_web):
+    """
+    Submit a claim with an invalid kid and check that it is rejected.
```
plietar, line 145: At some point which should also check for invalid UTF8 in the COSE bstr kid (which aren't representable in DID/JWK).

plietar (approved):

Issue #53, opened 2022-12-20
Define service DID document and add `did:web` endpoint


letmaik:
The goal is to experiment with representing the service's receipt trust anchors across disaster recoveries through a stable DID.

#25 was a first step towards that goal which allows setting the service DID and including it in receipts. The next step is to define what the corresponding DID document would look like when using did:web and exposing it under `GET <domain>/.well-known/did.json`. 

There are some challenges that need to be solved:

## Usage of JWK
Verification material in DID documents is expressed using one or more JWKs (JSON Web Key), where each JWK represents a public key. The expectation is that the public key verifies the signature of what was signed.

In CCF, there is a hierarchy of service key and node key, both wrapped in certificates. The node key differs per node and is the one that signs the tree root and hence is used to verify receipt signatures. The service key only changes when a disaster recovery event occurs.

There are two options:

Option A: Node public keys could be exposed in the DID document which would make most sense from a DID point of view but is not economical as the list will grow too large very quickly.

Option B: Service public keys could be exposed in the DID document which makes more sense from a trust anchor and efficiency point of view but doesn't seem to be immediately compatible with existing DID practices.

To make Option B work, the details of CCF including the indirection through the node certificate could be hidden behind the receipt/tree algorithm, whose interface would receive the service key and internally uses the node key and node certificate for verification after checking that the node certificate was signed by the service key.
 
## Validation using historic keys

Receipts shall always be verifiable, even if they are old and the current signing keys of a service are not the ones used in the old receipts.

From the point of CCF, old service keys can be considered as valid only for the range of transactions for which they were in use. The current service key has no upper bound, until a disaster recovery happens.

In DID, the only standard way to restrict the use of a key to validate new signatures is by either including or not including it in the DID document. If a DID method supports retrieving old versions of DID documents via `versionId` or `versionTime` and a signature includes that metadata and there is a reliable way to establish the point in time at which the signature was made, then resolving an old DID document state would allow to retrieve the correct key and verify the signature. `did:web` does not support this.

Option A: Include all historic keys and add non-standard metadata to each key which restricts its use. This could be a transaction sequence range, though currently the sequence numbers are not exposed in receipts. It could also be a time range, though CCF does not record time in general.

Option B: ?

plietar:
I had a stab at this, and unfortunately it seems CCF does not currently allow us to use the `/.well-known` path.

[...]

PR #68, opened 2023-01-11
Add a DID-based service identifier to the ledger.

plietar:
A DID document is hosted at /scitt/did.json, allowing the service identifier to be resolved through did:web.

For now, the document only includes the current service certificate, but in the future this will be included to include past certificates too, allowing historical receipts to be verified.

Related to #53

### DIFF
--- a/pyscitt/pyscitt/did.py
+++ b/pyscitt/pyscitt/did.py
@@ -4,6 +4,8 @@
 from typing import Optional
 from urllib.parse import quote, unquote
 
+import httpx
+
 from .crypto import Pem, Signer
 
 DID_FILENAME = "did.json"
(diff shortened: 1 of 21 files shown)


pyscitt/pyscitt/did.py
```diff
@@ -75,3 +78,20 @@ def get_signer(private_key: Pem, did_doc: dict, kid: Optional[str] = None) -> Si
...
     return Signer(private_key, did_doc["id"], kid, algorithm)
+
+
+class Resolver:
+    def resolve(self, did: str) -> dict:
+        parts = did.split(":", 3)
+        if len(parts) < 3:
+            raise ValueError("Malformed DID")
+
+        if parts[0] != "did" or parts[0] != "web":
```
letmaik, line 89:
```suggestion
        if parts[0] != "did" or parts[1] != "web":
```

letmaik (approved):
Looks good. This only surfaces the current service identity in the DID document which means receipts from previous service identities can't be validated. We should keep track of this and maybe implement that for the version next month.

letmaik:
Can you link this to an issue?

Issue #50, opened 2022-12-19
Monitor performance


letmaik:
We should define some performance metrics and monitor them similar to how CCF does in their repository.

Metrics could be:
- Throughput on POST `/entries`
  - did:web using cached resolution
  - X.509
- Throughput on GET `/entries/<txid>/receipt`
  - Cached historical transaction
- Latency as time to receipt following POST `/entries` 
  - Uncached did:web, requires at minimum two calls to POST `/entries`, followed by one or more GET `/entries/<txid>/receipt` until receipt is ready
  - Cached did:web, requires one call to POST `/entries`, followed by one or more GET `/entries/<txid>/receipt`
  - X.509, same as cached did:web
  - Note: Measuring latency is tricky as it relies on some parameters like CCF's tree signing interval (configurable) and the client delay between retries both for POST `/entries` in the case of uncached did:web and for receipt fetching.
- Memory usage after inserting a large number of entries
  - Single issuer
  - Many issuers
  - With/without prefix tree

Using a small claim payload size for now is probably fine, like 1 KB.

achamayou:
Dupe of #208, now done.

PR #69, opened 2023-01-11
Add latency perf test

letmaik:
This implements some of #50, in particular the latency for `POST /entries` as well as time-to-receipt which involves further GET calls. It is implemented as a pytest that's disabled by default. It writes a `perf.json` to the current folder (added to `.gitignore`) which gets uploaded as pipeline artifact in CI.

### DIFF
--- a/.pipelines/azure-pipelines.yml
+++ b/.pipelines/azure-pipelines.yml
@@ -38,6 +38,8 @@ stages:
           Pool: 1es-dcv3
           Environment:
             ENABLE_PREFIX_TREE: OFF
+          FunctionalTestArguments: '--enable-perf'
+          PublishPerfResults: true
 
       - template: common.yml
         parameters:
--- a/.pipelines/common.yml
+++ b/.pipelines/common.yml
@@ -84,5 +84,5 @@ jobs:
       - ${{ if parameters.PublishPerfResults }}:
         - task: PublishPipelineArtifact@1
           inputs:
-            artifactName: "perf-results"
+            artifactName: "${{ parameters.Name }}-perf-results"
             targetPath: $(Build.SourcesDirectory)/perf.json
--- a/.pipelines/common.yml
+++ b/.pipelines/common.yml
@@ -22,6 +22,9 @@ parameters:
   - name: FunctionalTestArguments
     type: string
     default: ''
+  - name: PublishPerfResults
+    type: boolean
+    default: false
 jobs:
   - job: ${{ parameters.Name }}
     displayName: ${{ parameters.DisplayName }}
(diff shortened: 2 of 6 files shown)


.pipelines/azure-pipelines.yml
```diff
@@ -38,6 +38,8 @@ stages:
           Pool: 1es-dcv3
           Environment:
             ENABLE_PREFIX_TREE: OFF
+          FunctionalTestArguments: '--enable-perf'
```
letmaik, line 41: If this slows down CI too much we may also introduce a separate pipeline that runs on commits to `main`, not in PRs.

pyscitt/pyscitt/client.py
```diff
@@ -153,6 +153,7 @@ def request(
         url,
         *,
         retry_on=[],
+        wait_time=None,
```
plietar, line 156: What about making this a property of the Client instead? You can easily create a "modified" client in tests by taking the `client` fixture and calling `client.replace(wait_time=0)`.

See the authentication tests, where it gets used to set credentials: https://github.com/microsoft/scitt-ccf-ledger/blob/566de22714d6ebc7745a8582a617da2ef5d832a8/test/test_auth.py#L50

test/test_perf.py
```diff
@@ -0,0 +1,83 @@
...
+        # the CCF network and is hence just a rough estimate.
+        stats = {
+            "latency_did_web_submit_s": latency_did_web_submit_s,
+            "latency_did_web_submit_and_receipt_s": latency_did_web_submit_and_receipt_s,
+            "latency_x5c_submit_s": latency_x5c_submit_s,
+            "latency_x5c_submit_and_receipt_s": latency_x5c_submit_and_receipt_s,
+        }
+
+        # Write stats to file for further analysis.
+        with open("perf.json", "w", encoding="utf-8") as fp:
```
plietar, line 82: This is probably fine for now, but eventually I think we'll want some pattern for reporting metrics from multiple different tests.

.pipelines/common.yml
```diff
@@ -77,3 +80,9 @@ jobs:
 
       - script: ./run_functional_tests.sh ${{ parameters.FunctionalTestArguments }}
         displayName: Run Functional Tests
+
+      - ${{ if parameters.PublishPerfResults }}:
+        - task: PublishPipelineArtifact@1
+          inputs:
+            artifactName: "perf-results"
```
plietar, line 87: Looking at the DevOps page, it seems all the jobs in the pipeline are writing to the same artifact name and overwriting each other. I think the `perf-results` name is going to need to be parametrized.

Maybe this works?
```suggestion
            artifactName: "${{ parameters.Name }}-perf-results"
```

plietar (changes requested):
This is probably going to suffer from the Nagle / delayed ACK issue I've seen before. The proper way to fix it would probably be to make a feature request on httpx.

.pipelines/common.yml
letmaik: Yeah, for now there's only a single job that runs the perf test, but you're right, we'll probably extend that to more jobs in the future.

plietar: Oh I hadn't even realized that. I wonder what happens if multiple jobs write the same artifact then? Does it throw an error?

letmaik: Yes, you can't override.

plietar (approved):

PR #73, opened 2023-01-13
WIP: Add support for Notary claims

darracott:
This PR allows claims generated by `notation` to be submitted to SCITT. 

To do:
- Tidy up (in particular the validation of header parameters for vanilla SCITT claims and Notary claims).
- Add a test cases submitting a Notary RSA signed claim and a Notary ECDSA signed claim.

Note:
- t_cose does not currently support custom header parameters in `crit`. This PR brings back a lot of code to work around that limitation so that we can verify Notary claims' signatures without using t_cose. 
  -  Once t_cose allows custom parameters in `crit` then we should remove the workaround code which is marked by comments stating the code is "Temporarily needed for notary_verify().".


app/src/cose.h
```diff
@@ -207,6 +439,388 @@ namespace scitt::cose
...
+    if (key_alg.has_value() && header_alg != key_alg.value())
+    {
+      throw COSESignatureValidationError(
+        "Algorithm mismatch between protected header and public key");
+    }
+
+    auto md_type = get_md_type(header_alg);
+    auto ossl_md_type = get_openssl_md_type(md_type);
+    auto tbs = create_tbs(cose_sign1);
+    auto signature = get_signature(cose_sign1);
```
plietar, line 775: There's already a `extract_sign1_fields` method in this file you can use.

app/src/cose.h
```diff
@@ -207,6 +439,388 @@ namespace scitt::cose
...
+    if (key_len_bits % 8)
+    {
+      key_len_bytes++;
+    }
+
+    return key_len_bytes;
+  }
+
+  // Temporarily needed for notary_verify().
+  std::vector<uint8_t> create_tbs(const std::vector<uint8_t>& cose_sign1)
```
plietar, line 561: If you use `extract_sign1_fields` in the caller (which is needed to get the signature), you can change this
```suggestion
  std::vector<uint8_t> create_tbs(std::span<const uint8_t> protected, std::span<const uint8_t> payload)
```
and not need to do the decoding here.

app/src/cose.h
```diff
@@ -207,6 +439,388 @@ namespace scitt::cose
...
+    err = QCBOREncode_Finish(&encode_ctx, &encoded_cbor);
+    if (err != QCBOR_SUCCESS)
+    {
+      throw std::runtime_error("Error encoding CBOR");
+    }
+
+    auto cbor = std::vector<uint8_t>(
+      (uint8_t*)encoded_cbor.ptr,
+      (uint8_t*)encoded_cbor.ptr + encoded_cbor.len);
+    return cbor;
```
plietar, line 642: Use the `cbor::encoder` class, which hides some of the buffer management away.
```c++
    cbor::encoder ctx(cose_sign1.size() + 1024);
    QCBOREncode_OpenArray(ctx);
    [...]
    QCBOREncode_CloseArray(ctx);
    return ctx.finish();
```

app/src/cose.h
```diff
@@ -207,6 +439,388 @@ namespace scitt::cose
...
+
+    /* Calculation of size per RFC 8152 section 8.1 -- round up to
+     * number of bytes. */
+    key_len_bytes = (unsigned)key_len_bits / 8;
+    if (key_len_bits % 8)
+    {
+      key_len_bytes++;
+    }
+
+    return key_len_bytes;
```
plietar, line 557: I'm pretty sure this is equivalent.
```suggestion
    return (key_len_bits + 7) / 8;
```

app/src/cose.h
```diff
@@ -207,6 +439,388 @@ namespace scitt::cose
...
+      throw std::runtime_error("Error encoding CBOR");
+    }
+
+    auto cbor = std::vector<uint8_t>(
+      (uint8_t*)encoded_cbor.ptr,
+      (uint8_t*)encoded_cbor.ptr + encoded_cbor.len);
+    return cbor;
+  }
+
+  static enum t_cose_err_t ecdsa_signature_cose_to_der(
```
plietar, line 645: I think we're better off copying from CCF than t_cose for this: https://github.com/microsoft/CCF/blob/5972983bce5e345666e5bf87d0f28ac1f7f05a84/src/crypto/ecdsa.cpp#L14-L45

Annoyingly, this API is public in CCF 4.x but not 3.x.

app/src/verifier.h
```diff
@@ -130,15 +132,243 @@ namespace scitt::verifier
...
+    {
+      // Validate protected header
+      validate_notary_protected_header(phdr, configuration);
 
+      // Notary puts the x5chain in unprotected header,
+      // hence decode unprotected header.
+      cose::UnprotectedHeader unphdr;
+      try
+      {
+        unphdr = cose::decode_unprotected_header(data);
```
plietar, line 288: We should decode the protected and unprotected headers at the same time IMO (it's fine to keep them as separate structs).
Have one `decode_headers(std::span<const uint8_t> cose_sign1)` function, which calls both `decode_protected_header(QCBORDecodeContext* ctx)` and `decode_unprotected_header(QCBORDecodeContext* ctx)`.

app/src/cose.h
```diff
@@ -207,6 +439,388 @@ namespace scitt::cose
     {}
   };
 
+  // Temporarily needed for notary_verify().
+  std::vector<uint8_t> qcbor_buf_to_vector(const UsefulBufC& buf)
```
plietar, line 443: `cbor::as_vector` does the same thing already.

app/src/cose.h
```diff
@@ -131,25 +300,67 @@ namespace scitt::cose
...
     }
+    if (header_items[CRIT_INDEX].uDataType != QCBOR_TYPE_NONE)
+    {
+      parsed.crit = std::vector<std::variant<int64_t, std::string>>();
+      QCBORItem critItem = header_items[CRIT_INDEX];
+      if (critItem.uDataType != QCBOR_TYPE_ARRAY)
+      {
+        throw COSEDecodeError(
+          "Value type of crit in protected header is not of type array");
+      }
```
plietar, line 323: `t_cose` already checks this for us, since we specified `uDataType = QCBOR_TYPE_ARRAY`.

app/src/cose.h
```diff
@@ -27,30 +27,161 @@
...
+    QCBORDecode_GetItemsInMap(&ctx, header_items);
+
+    qcbor_result = QCBORDecode_GetError(&ctx);
+    if (qcbor_result != QCBOR_SUCCESS)
+    {
+      throw COSEDecodeError("Failed to decode unprotected header");
+    }
+    if (header_items[X5CHAIN_INDEX].uDataType != QCBOR_TYPE_NONE)
+    {
+      parsed.x5chain = std::vector<std::vector<uint8_t>>();
```
plietar, line 118: This code already exists in `decode_protected_header`. It should be refactored into a common function (it's fine if the exception messages don't mention protected/unprotected).

app/src/cose.h
```diff
@@ -27,30 +27,161 @@
...
+      else
+      {
+        CCF_APP_INFO("x5c type", chainItem.uDataType);
+        throw COSEDecodeError(
+          "Value type of x5chain in unprotected header is not array or byte "
+          "string.");
+      }
+    }
+    QCBORDecode_ExitMap(&ctx);
+    QCBORDecode_ExitBstrWrapped(&ctx);
```
plietar, line 158: This seems wrong. There's no bstr wrapping here.

Also we need to check for errors on the context by calling `QCBORDecode_Finish()` (this is also missing from `decode_protected_header` actually).

app/src/cose.h
```diff
@@ -27,30 +27,161 @@
...
+
+    uint64_t tag = QCBORDecode_GetNthTagOfLast(&ctx, 0);
+    if (tag != CBOR_TAG_COSE_SIGN1)
+    {
+      throw COSEDecodeError("COSE_Sign1 is not tagged");
+    }
+    QCBORItem coseItem;
+    QCBORDecode_GetNext(&ctx, &coseItem);
+
+    struct q_useful_buf_c unprotected_parameters;
```
plietar, line 94: This isn't used anywhere.

app/src/cose.h
```diff
@@ -131,25 +300,67 @@ namespace scitt::cose
...
+          "Value type of crit in protected header is not of type array");
+      }
+      int lenOfArray = critItem.val.uCount;
+      if (lenOfArray == 0)
+      {
+        throw COSEDecodeError(
+          "Cannot have crit array of length 0 in cose protected header.");
+      }
+      QCBORDecode_EnterArrayFromMapN(&ctx, COSE_HEADER_PARAM_CRIT);
+      for (int i = 0; i < lenOfArray; i++)
```
plietar, line 331: This is going to behave pretty badly if someone uses a indefinite length array. In that case, `lenOfArray` is `UINT16_MAX`. It is better to call `GetNext` repeatedly, until it returns `QCBOR_ERR_NO_MORE_ITEMS`.
See https://github.com/laurencelundblade/t_cose/blob/f9859dd1525623cd9de7d23ac160acec7a859ef1/src/t_cose_parameters.c#L138-L184

The `lenOfArray == 0` check should also happen against the resulting vector.

Same applies for the x5chain code.

app/src/cose.h
```diff
@@ -131,25 +300,67 @@ namespace scitt::cose
...
+      {
+        QCBORDecode_GetNext(&ctx, &critItem);
+        if (critItem.uDataType == QCBOR_TYPE_TEXT_STRING)
+        {
+          parsed.crit->push_back(std::variant<int64_t, std::string>(
+            std::string(cbor::as_string(critItem.val.string))));
```
plietar, line 337:
```suggestion
          parsed.crit->push_back(std::string(cbor::as_string(critItem.val.string))));
```

I don't think this needs to be so explicit.

app/src/cose.h
```diff
@@ -131,25 +300,67 @@ namespace scitt::cose
...
+        }
+        else if (critItem.uDataType == QCBOR_TYPE_INT64)
+        {
+          parsed.crit->push_back(
+            std::variant<int64_t, std::string>(critItem.val.int64));
```
plietar, line 342:
```suggestion
          parsed.crit->push_back(critItem.val.int64);
```

app/src/verifier.h
```diff
@@ -130,15 +132,243 @@ namespace scitt::verifier
...
+        {
+          throw cose::COSEDecodeError(
+            "Missing io.cncf.notary.signingTime in protected header");
+        }
+        if (notary_authentic_signing_time.has_value())
+        {
+          throw cose::COSEDecodeError(
+            "io.cncf.notary.authenticSigningTime not allowed in protected "
+            "header when io.cncf.notary.signingScheme is `notary.x509`");
+        }
```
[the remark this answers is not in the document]
letmaik: Spec says "This claim is REQUIRED and only valid when signing scheme is notary.x509.signingAuthority ."

app/src/verifier.h
```diff
@@ -130,15 +132,243 @@ namespace scitt::verifier
...
+      if (
+        std::find(
+          crit.begin(),
+          crit.end(),
+          std::variant<int64_t, std::string>(
+            std::string("io.cncf.notary.signingScheme"))) == crit.end())
+      {
+        throw cose::COSEDecodeError(
+          "crit must contain 'io.cncf.notary.signingScheme'");
+      }
```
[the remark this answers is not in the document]
letmaik: I added `is_critical` to `ProtectedHeaders`.

app/src/cose.h
```diff
@@ -121,92 +220,474 @@ namespace scitt::cose
...
+      case crypto::MDType::SHA512:
+        return EVP_sha512();
+      default:
+        throw std::runtime_error("Unsupported hash algorithm");
+    }
+    return nullptr;
+  }
+
+  // Temporarily needed for notary_verify().
+  static unsigned ecdsa_key_size(EVP_PKEY* key_evp)
```
plietar, line 500: This isn't being used anymore

plietar (approved):

app/src/cose.h
```diff
@@ -396,17 +804,27 @@ namespace scitt::cose
...
 
-    QCBOREncode_AddBytes(encoder, protected_header);
+    QCBOREncode_AddBytes(encoder, cbor::from_bytes(protected_header));
 
     // unprotected header
     QCBOREncode_OpenMap(encoder);
     QCBOREncode_OpenArrayInMapN(encoder, COSE_HEADER_PARAM_SCITT_RECEIPTS);
     QCBOREncode_AddEncoded(encoder, cbor::from_bytes(receipt));
     QCBOREncode_CloseArray(encoder);
+    if (x5chain.has_value())
```
plietar, line 814: If `x5chain` only has one element it should be serialized as a single bstr, not an array.

https://datatracker.ietf.org/doc/html/draft-ietf-cose-x509-09 Section 2
> If a single certificate is conveyed, it is placed in a CBOR byte string.
         
And in Table 1: `COSE_X509 = bstr / [ 2*certs: bstr ]`, where `2*` means it needs at least two elements.

app/src/verifier.h
```diff
@@ -66,76 +58,322 @@ namespace scitt::verifier
...
+      kv::Tx& tx,
+      const Configuration& configuration)
+    {
+      // Validate protected header
+      validate_notary_protected_header(phdr, configuration);
+
+      // Validate unprotected header
+      // Unprotected header must contain x5chain else the notary claim cannot be
+      // verified.
+      if (!uhdr.x5chain.has_value())
```
plietar, line 279: [The Notary COSE spec](https://github.com/notaryproject/notaryproject/blob/900a36df30315803e221bd45754a53395fae359b/specs/signature-envelope-cose.md#unprotected-headers) says `Optionally, this header can be presented in the protected header`, which suggests we should actually look at both headers. Probably we should be looking at the protected one first, or even fail if both are present.

(oddly enough, the JWS spec has no such phrasing, and only allows it in the unprotected header)

app/src/verifier.h
```diff
@@ -66,76 +58,322 @@ namespace scitt::verifier
...
+          "Notary claim's unprotected header is missing an x5chain (label 33) "
+          "parameter.");
+      }
+
+      PublicKey key;
+      // Verify the chain of certs against the x509 root store.
+      auto roots = x509_root_store(tx);
+      auto cert = verify_chain(roots, uhdr.x5chain.value());
+      key = PublicKey(cert, std::nullopt);
+      return key;
```
plietar, line 291: The assignment to a variable defined 3 lines up is a bit confusing and unnecessary IMO.
```
return PublicKey(cert, std::nullopt);
```

test/test_ccf.py
```diff
@@ -12,6 +18,23 @@
 from pyscitt.verify import verify_receipt
 
 
+# Temporary monkey-patch for pycose until https://github.com/TimothyClaeys/pycose/pull/107
+# is released.
```
plietar, line 22: Can this be removed now you've pinned to a git commit?

pyscitt/pyscitt/prefix_tree.py
```diff
@@ -110,8 +110,9 @@ def verify(
         self, claim: Union[Sign1Message, bytes], service_params: ServiceParameters
     ):
         if isinstance(claim, bytes):
-            claim = CoseMessage.decode(claim)
-        assert isinstance(claim, Sign1Message)
+            claim_ = CoseMessage.decode(claim)
+            assert isinstance(claim_, Sign1Message)
+            claim = claim_
```
plietar, line 115: `Sign1Message.decode(claim)` should have the right type, thanks to https://github.com/TimothyClaeys/pycose/pull/102, which means you shouldn't need the `assert` and temporary variable.

app/src/cose.h
```diff
@@ -20,87 +20,186 @@
...
+    ProtectedHeader parsed;
+
+    // Adapted from parse_cose_header_parameters in t_cose_parameters.c.
+    // t_cose doesn't support custom header parameters yet.
+
+    QCBORError qcbor_result;
+
     struct q_useful_buf_c protected_parameters;
     QCBORDecode_EnterBstrWrapped(
       &ctx, QCBOR_TAG_REQUIREMENT_NOT_A_TAG, &protected_parameters);
```
plietar, line 170: According to the QCBOR docs, we should be able to pass NULL as the third parameter and not need to define `protected_parameters` (since we don't use it anyway).

app/src/cose.h
```diff
@@ -121,92 +220,463 @@ namespace scitt::cose
...
+    }
+
+    uint64_t tag = QCBORDecode_GetNthTagOfLast(&ctx, 0);
+    if (tag != CBOR_TAG_COSE_SIGN1)
+    {
+      throw COSEDecodeError("COSE_Sign1 is not tagged");
+    }
+
+    auto phdr = decode_protected_header(ctx);
+    auto uhdr = decode_unprotected_header(ctx);
```
plietar, line 431: I think it would be good to do a pair of `QCBORDecode_GetByteString`, a `QCBORDecode_CloseArray` and finally a `QCBORDecode_Finish` with an error check here, to ensure the enveloppe is well-formed.

Admittedly t_cose is gonna be doing this eventually, but since this is the first piece of code to be decoding the message it needs to be reasonably robust to garbage.

app/src/cose.h
```diff
@@ -121,92 +220,463 @@ namespace scitt::cose
...
+    // EdDSA does not support pre-hashed messages as input.
+
+    // Sig_structure = [
+    //     context: "Signature1",
+    //     body_protected: empty_or_serialized_map,
+    //     external_aad: bstr,
+    //     payload: bstr
+    // ]
+
+    cbor::encoder ctx(cose_size + 1024);
```
plietar, line 515: `protected_header.size() + payload.size() + 1024` seems good enough (if anything it's more accurate), no need to have a `cose_size` parameter.

app/src/cose.h
```diff
@@ -121,92 +220,463 @@ namespace scitt::cose
...
+    QCBORDecodeContext ctx;
+    QCBORDecode_Init(
+      &ctx, cbor::from_bytes(cose_sign1), QCBOR_DECODE_MODE_NORMAL);
+
+    QCBORDecode_EnterArray(&ctx, nullptr);
+
+    QCBORItem item;
+
+    // protected headers
+    QCBORDecode_VGetNext(&ctx, &item);
```
plietar, line 596: All three `QCBORDecode_VGetNext` calls need to be replaced with `QCBORDecode_GetByteString`. Otherwise someone could put another type of element (eg. an integer), and create a type confusion when loading `item.val.string`.

(Using `QCBORDecode_VGetNextConsume` for the unprotected headers is fine IMO)

I realise this is old code and I even think I wrote this, but only just notice. It's much more important to fix this now this is running on envelopes that haven't been checked by t_cose.

app/src/cose.h
```diff
@@ -20,87 +20,186 @@
...
   struct COSEDecodeError : public std::runtime_error
   {
     COSEDecodeError(const std::string& msg) : std::runtime_error(msg) {}
   };
 
   struct ProtectedHeader
   {
+    // All headers are optional here but optionality will later be validated
+    // according to the COSE profile of the claim.
+    std::optional<std::string> profile;
```
plietar, line 65: I'm not sure I understand why we added this, given it's not being used anywhere (and we infer the profile from the presence of header parameters)

pyscitt/setup.py
```diff
@@ -17,7 +17,8 @@
         "cryptography==38.*",  # needs to match ccf
         "httpx",
         "cbor2",
-        "pycose>=1.0.1",
+        # TODO: remove this once pycose >= 1.0.2 is released
+        "pycose @ git+https://github.com/TimothyClaeys/pycose@94db358eda640966c0e0e9148110b6c66763f9e5#egg=pycose",
```
plietar, line 21: There's an xfail marker on `test_encoding.py`'s `test_verify_receipt` that can be removed now.

app/src/cose.h
darracott: Done in https://github.com/microsoft/scitt-ccf-ledger/pull/90

app/src/verifier.h
```diff
@@ -66,76 +58,322 @@ namespace scitt::verifier
...
-      PublicKey key;
-      if (x5chain.has_value())
+      std::optional<std::string> assertion_method_id;
+      if (!issuer.has_value())
       {
-        // Verify the chain of certs against the x509 root store.
-        auto roots = x509_root_store(tx);
-        auto cert = verify_chain(roots, x5chain.value());
-        key = PublicKey(cert, std::nullopt);
+        throw VerificationError(
```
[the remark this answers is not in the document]
darracott: Remove this check in https://github.com/microsoft/scitt-ccf-ledger/pull/90

app/src/cose.h
```diff
@@ -20,87 +20,186 @@
...
+          parsed.push_back(cbor::as_vector(item.val.string));
+        }
+        else
+        {
+          throw COSEDecodeError(
+            "Next item in x5chain was not of type byte string.");
+        }
+      }
+      QCBORDecode_ExitArray(&ctx);
+      if (parsed.empty())
```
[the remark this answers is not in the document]
darracott: Added a log in https://github.com/microsoft/scitt-ccf-ledger/pull/90 so we can see when claims get this wrong but currently we know implementations have missed this subtlety of the spec and so we won't reject claims that encode a single cert inside an array.

app/src/verifier.h
darracott: Done in https://github.com/microsoft/scitt-ccf-ledger/pull/90

app/src/cose.h
darracott: Done in https://github.com/microsoft/scitt-ccf-ledger/pull/90

app/src/cose.h
darracott: Done in https://github.com/microsoft/scitt-ccf-ledger/pull/90

app/src/cose.h
darracott: Done in https://github.com/microsoft/scitt-ccf-ledger/pull/90

app/src/cose.h
darracott: Done in https://github.com/microsoft/scitt-ccf-ledger/pull/90 but could you specifically review this bit please

app/src/cose.h
darracott: Removed it in https://github.com/microsoft/scitt-ccf-ledger/pull/90. - Originally it made the structure of the code more clear but now we've got 3 profiles, after splitting the vanilla case into IETF and X509, I think the structure is clear enough without it.

pyscitt/setup.py
darracott: Removed in https://github.com/microsoft/scitt-ccf-ledger/pull/90

PR #76, opened 2023-01-16
Reduce retry-after from 3 to 1 second

### LINKED ISSUE #70: Reduce `retry-after` header value
We currently use 3 seconds as retry-after value sent back to the client when a historical query is not cached in-memory yet. This seems excessive and should be lowered to 1 second.

letmaik:
Closes #70.

### DIFF
--- a/app/src/historical/historical_queries_adapter.h
+++ b/app/src/historical/historical_queries_adapter.h
@@ -123,7 +123,7 @@ namespace scitt::historical
       if (historical_state == nullptr)
       {
         args.rpc_ctx->set_response_status(HTTP_STATUS_ACCEPTED);
-        constexpr size_t retry_after_seconds = 3;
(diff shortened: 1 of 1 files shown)


plietar:
Do you know if this code path covers both cases, where a receipt does not exist yet and the case where it needs to be loaded from disk?

plietar (approved):

letmaik:
It only covers the case where it needs to be loaded from disk. The other case where the transaction doesn't exist yet returns a 404 with error code "TransactionPendingOrUnknown" without retry-after. Because we don't know whether the requested transaction will ever come into existence a 404 is returned, and for that it doesn't really make sense to return retry-after. It's all a bit unfortunate but I think the semantics of CCF don't allow to do anything more meaningful without optimistically guessing that the tx id was correct and will in fact become available soon.

Issue #72, opened 2023-01-13
Improved logging


letmaik:
The CCF app currently only logs the following cases:
- A non-2xx response is sent back to the user. Note that this does not apply to the historical query adapter.
- A did:web resolution is triggered.
- The DID document for an issuer is updated.

While that's a good start, a bit more should be logged:
- Non-2xx responses created by the historical query adapter.
- 202 responses when historical state is not loaded yet
- Details of the entry being stored as part of POST /entries
  - If possible, the transaction id should be logged as well to allow correlation with follow-up requests.
- 200 responses of all endpoints

Overall, for a given request there should always be a single log entry that logs the request/response itself (with HTTP status code, request path and query parameters, possible error details). Additional log entries may add further details or diagnostics but should be correlatable via some identifier. 

Related: #74.

PR #79, opened 2023-01-16
Add tracing adapter

### LINKED ISSUE #74: Add trace correlation headers
Tracing of requests across systems typically relies on HTTP headers to establish correlation ids.

There's https://www.w3.org/TR/trace-context/ which is seeing adoption but Microsoft historically also uses `x-ms-client-request-id` and `x-ms-request-id`.

For compatibility, we can start with `x-ms-request-id`/`x-ms-client-request-id` and include those in log messages. Later on we can add W3C Trace Context headers and eventually deprecate and then remove the `x-ms-*` headers.

letmaik:
Closes #74 and #72.

TODO:
- Create issue to allow extending tracing adapter log messages (e.g. for attaching error code instead of logging a separate message; similarly for logging the TxId)

### DIFF
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -285,6 +285,9 @@ namespace scitt
         });
 
         ctx.rpc_ctx->set_response_status(HTTP_STATUS_CREATED);
+
+        // TODO log CoseProfile
+        SCITT_INFO("ClaimSizeKb={}", body.size() / 1024);
       };
 
       make_endpoint(
@@ -672,7 +675,7 @@ namespace scitt
           std::move(resolution.resolution_metadata);
 
         issuers->put(issuer, issuer_info.value());
-        SCITT_INFO("Updated DID document for issuer {}", issuer);
+        SCITT_INFO("Updated DID document for {}", issuer);
 
         return ccf::make_success();
       };
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -151,6 +151,25 @@ namespace scitt
       return std::nullopt;
     }
 
+    ccf::endpoints::Endpoint make_endpoint(
+      const std::string& method,
+      ccf::RESTVerb verb,
+      const ccf::endpoints::EndpointFunction& f,
+      const ccf::AuthnPolicies& ap) override
+    {
+      std::function<ccf::ApiResult(timespec & time)> get_time =
+        [this](timespec& time) {
+          return this->get_untrusted_host_time_v1(time);
+        };
+
+      return ccf::UserEndpointRegistry::make_endpoint_with_local_commit_handler(
+        method,
+        verb,
+        tracing_adapter(error_adapter(f), method, get_time),
+        tracing_local_commit_callback,
+        ap);
+    }
+
   public:
     AppEndpoints(ccfapp::AbstractNodeContext& context_) :
       ccf::UserEndpointRegistry(context_)
@@ -179,11 +198,6 @@ namespace scitt
       {
         srand(untrusted_host_time.tv_sec);
       }
-
-      std::function<ccf::ApiResult(timespec & time)> get_time =
-        [this](timespec& time) {
-          return this->get_untrusted_host_time_v1(time);
-        };
 
       static constexpr auto post_entry_path = "/entries";
       auto post_entry = [this](EndpointContext& ctx) {
@@ -290,12 +304,7 @@ namespace scitt
         SCITT_INFO("ClaimSizeKb={}", body.size() / 1024);
       };
 
-      make_endpoint_with_local_commit_handler(
-        post_entry_path,
-        HTTP_POST,
-        tracing_adapter(error_adapter(post_entry), get_time),
-        tracing_local_commit_callback,
-        authn_policy)
+      make_endpoint(post_entry_path, HTTP_POST, post_entry, authn_policy)
         .install();
 
       auto is_tx_committed =
@@ -371,13 +380,8 @@ namespace scitt
       make_endpoint(
         get_entry_path,
         HTTP_GET,
-        tracing_adapter(
-          error_adapter(scitt::historical::adapter(
-            get_entry,
-            state_cache,
-            is_tx_committed,
-            get_tx_id_from_request_path)),
-          get_time),
+        scitt::historical::adapter(
+          get_entry, state_cache, is_tx_committed, get_tx_id_from_request_path),
         authn_policy)
         .set_forwarding_required(ccf::endpoints::ForwardingRequired::Never)
         .install();
@@ -422,13 +426,11 @@ namespace scitt
       make_endpoint(
         get_entry_receipt_path,
         HTTP_GET,
-        tracing_adapter(
-          error_adapter(scitt::historical::adapter(
-            get_entry_receipt,
-            state_cache,
-            is_tx_committed,
-            get_tx_id_from_request_path)),
-          get_time),
+        scitt::historical::adapter(
+          get_entry_receipt,
+          state_cache,
+          is_tx_committed,
+          get_tx_id_from_request_path),
         authn_policy)
         .set_forwarding_required(ccf::endpoints::ForwardingRequired::Never)
         .install();
@@ -552,8 +554,7 @@ namespace scitt
       make_endpoint(
         get_entries_tx_ids_path,
         HTTP_GET,
-        tracing_adapter(
-          error_adapter(ccf::json_adapter(get_entries_tx_ids)), get_time),
+        ccf::json_adapter(get_entries_tx_ids),
         authn_policy)
         .set_auto_schema<void, GetEntriesTransactionIds::Out>()
         .add_query_parameter<size_t>(
@@ -578,8 +579,7 @@ namespace scitt
       make_endpoint(
         get_issuers_path,
         HTTP_GET,
-        tracing_adapter(
-          error_adapter(ccf::json_adapter(get_issuers)), get_time),
+        ccf::json_adapter(get_issuers),
         authn_policy)
         .install();
 
@@ -605,8 +605,7 @@ namespace scitt
       make_endpoint(
         get_issuer_info_path,
         HTTP_GET,
-        tracing_adapter(
-          error_adapter(ccf::json_adapter(get_issuer_info)), get_time),
+        ccf::json_adapter(get_issuer_info),
         authn_policy)
         .install();
 
@@ -681,12 +680,10 @@ namespace scitt
         return ccf::make_success();
       };
 
-      make_endpoint_with_local_commit_handler(
+      make_endpoint(
         update_did_doc_path,
         HTTP_POST,
-        tracing_adapter(
-          error_adapter(ccf::json_adapter(update_did_doc)), get_time),
-        tracing_local_commit_callback,
+        ccf::json_adapter(update_did_doc),
         no_authn_policy)
         .set_auto_schema<PostDidResolution::In, void>()
         .install();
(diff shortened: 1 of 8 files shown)


app/src/tracing.h
```diff
@@ -0,0 +1,141 @@
...
+        + (time1.tv_nsec - time0.tv_nsec) / 1000000;
+  }
+
+  std::string create_request_id() {
+    std::stringstream stream;
+    stream << std::hex << rand();
+    return stream.str();
+  }
+
+  inline void SCITT_INFO(const std::string& msg)
```
plietar, line 36: We'll want these to be macros, which allows `__FILE__` and `__LINE__` to be accurate, and to get formatting syntax `SCITT_INFO("value={}", value)` working.

Unless [`source_location`](https://en.cppreference.com/w/cpp/utility/source_location) works on our compiler and we are willing to use CCF's logger directly to supply these, in which case we can get the formatting working using a variadic template.

app/src/tracing.h
```diff
@@ -0,0 +1,141 @@
...
+    return [fn, get_time](Ctx& ctx) {
+      request_id = create_request_id();
+      ctx.rpc_ctx->set_response_header(REQUEST_ID_HEADER, request_id);
+
+      client_request_id = ctx.rpc_ctx->get_request_header(CLIENT_REQUEST_ID_HEADER);
+
+      if (client_request_id.has_value())
+      {
+        std::smatch match;
+        if (!std::regex_match(client_request_id.value(), match, CLIENT_REQUEST_ID_REGEX))
```
plietar, line 67: Do we care? We never inspect the value, just forward it around, so it seems fine to allow anything.

app/src/tracing.h
```diff
@@ -0,0 +1,141 @@
...
+        msg
+        );
+  }
+
+  template <typename Fn, typename Ctx>
+  Fn generic_tracing_adapter(
+    Fn fn, const std::function<ccf::ApiResult(::timespec& time)>& get_time)
+  {
+    return [fn, get_time](Ctx& ctx) {
+      request_id = create_request_id();
```
plietar, line 59: I think we should clear this at the end of the function. A try/finally is probably the safest way to do so.

One pattern I'm imagining is we run an endpoint, variable gets set, endpoint terminates, we run indexing strategy code (which isn't part of a request) on the same thread, it inherits the leftover request ID and that gets logs, which would be pretty confusing.

app/src/tracing.h
```diff
@@ -0,0 +1,141 @@
...
+#include <regex>
+#include <sstream>
+#include <functional>
+#include <ccf/endpoint_context.h>
+#include <ccf/base_endpoint_registry.h>
+
+namespace scitt
+{
+  std::string REQUEST_ID_HEADER = "x-ms-request-id";
+  std::string CLIENT_REQUEST_ID_HEADER = "x-ms-client-request-id";
```
plietar, line 18:
```suggestion
  constexpr std::string_view REQUEST_ID_HEADER = "x-ms-request-id";
  constexpr std::string_view CLIENT_REQUEST_ID_HEADER = "x-ms-client-request-id";
```

`string_view` ensures they aren't mutable.

app/src/tracing.h
```diff
@@ -0,0 +1,141 @@
...
+
+  std::regex CLIENT_REQUEST_ID_REGEX("^[0-9a-zA-Z-]+");
```
plietar, line 20:
```suggestion
  const std::regex CLIENT_REQUEST_ID_REGEX("^[0-9a-zA-Z-]+");
```

app/src/tracing.h
ivarprudnikov, line 18: Azure SDK guidelines already advocate for `traceparent` and `tracestate` .
* https://azure.github.io/azure-sdk/general_azurecore.html#distributed-tracing-policy

API guidelines suggest followng the SDK guidelines but also mention common headers (implemented ATM) but it is mentioned like:

> In addition to distributed tracing, Azure also uses a set of common correlation headers

* https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#distributed-tracing--telemetry

ivarprudnikov: In another place [in the guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#http-request--response-pattern):

> The table below lists the headers most used by Azure services:
> ...
> _x-ms-request-id_ | Response 
> ...
> DO support all headers shown in _italics_

I suppose `x-ms-request-id` needs to be returned from the ledger ATM. But

app/src/tracing.h
letmaik: I added a comment, but yeah, doing some checking is useful as we insert this value directly in log messages and want to avoid misinterpretation of the log, e.g. if it contains whitespaces. It's a bit like SQL injection attacks.

app/src/tracing.h
letmaik: Done. Good call.

app/src/tracing.h
letmaik: Done.

app/src/main.cpp
```diff
@@ -272,10 +285,17 @@ namespace scitt
...
 
-      make_endpoint(
-        post_entry_path, HTTP_POST, error_adapter(post_entry), authn_policy)
+      make_endpoint_with_local_commit_handler(
+        post_entry_path,
+        HTTP_POST,
+        tracing_adapter(error_adapter(post_entry), get_time),
+        tracing_local_commit_callback,
+        authn_policy)
         .install();
```
plietar, line 299: At some point I think we should define our own `make_endpoint` function that provides all the right defaults. Happy to do this in a later PR.

app/src/tracing.h
```diff
@@ -0,0 +1,222 @@
...
+    std::optional<std::string> client_request_id;
+  };
+
+  void clear_trace_state()
+  {
+    request_id = "";
+    client_request_id = std::nullopt;
+  }
+
+  int diff_timespec_ms(
```
plietar, line 38: Should we have a higher resolution than ms? Or even better, use floating point seconds?

Because we are limited by CCF's 1ms ticking interval, we probably will never see anything between 0 and 1ms, but at least using a format with higher resolution will allow this to work if we get a better source of time in the future (maybe with SEV-SNP?).

app/src/tracing.h
```diff
@@ -0,0 +1,222 @@
...
+    return (time1.tv_sec - time0.tv_sec) * 1000 +
+      (time1.tv_nsec - time0.tv_nsec) / 1000000;
+  }
+
+  std::string create_request_id()
+  {
+    std::stringstream stream;
+    stream << std::hex << rand();
+    return stream.str();
+  }
```
plietar, line 50:
```suggestion
  std::string create_request_id()
  {
      return fmt::format("{:x}", rand());
  }
```

This might be a matter of personal preference, but I find this more readable.

app/src/tracing.h
```diff
@@ -0,0 +1,222 @@
...
+      request_id, \
+      ##__VA_ARGS__); \
+  else \
+    CCF_APP_INFO("RequestId={} " s, request_id, ##__VA_ARGS__)
+
+#define SCITT_FAIL(s, ...) \
+  if (client_request_id.has_value()) \
+    CCF_APP_FAIL( \
+      "ClientRequestId={} RequestId={} " s, \
+      client_request_id.value_or(""), \
```
plietar, line 66:
```suggestion
      client_request_id.value(), \
```

app/src/tracing.h
```diff
@@ -0,0 +1,222 @@
...
+      client_request_id =
+        ctx.rpc_ctx->get_request_header(CLIENT_REQUEST_ID_HEADER);
+
+      if (client_request_id.has_value())
+      {
+        // Validate client request id to avoid misinterpretation of the log,
+        // e.g. if it contains a space.
+        std::smatch match;
+        if (!std::regex_match(
+              client_request_id.value(), match, CLIENT_REQUEST_ID_REGEX))
```
plietar, line 91:
```suggestion
        if (!std::regex_match(client_request_id.value(), CLIENT_REQUEST_ID_REGEX))
```

Based on a quick glance at the docs, seems like you don't need a match object.

app/src/tracing.h
```diff
@@ -0,0 +1,222 @@
...
+            errors::InvalidInput,
+            "Invalid client request id.");
+          return;
+        }
+        ctx.rpc_ctx->set_response_header(
+          "x-ms-client-request-id", client_request_id.value());
+      }
+
+      ctx.rpc_ctx->set_user_data(
+        std::make_shared<AppData>(AppData{request_id, client_request_id}));
```
plietar, line 106: I think a comment would help, otherwise it's not totally obvious why we need this.
```suggestion
      // The user data is used to propagate the request IDs to the local commit callback
      ctx.rpc_ctx->set_user_data(
        std::make_shared<AppData>(AppData{request_id, client_request_id}));
```

app/src/tracing.h
```diff
@@ -0,0 +1,222 @@
...
+      }
+
+      auto duration_ms = diff_timespec_ms(start, end);
+
+      if (query.empty())
+      {
+        SCITT_INFO(
+          "Verb={} Path={} Status={} TimeMs={}",
+          ctx.rpc_ctx->get_request_verb().c_str(),
+          ctx.rpc_ctx->get_request_path(),
```
plietar, line 159: If a request includes path parameters, they are going to be included in this. This might make aggregation and analysis of logs difficult since we won't be able to group directly by path.

Maybe we should put the "method name", eg. `/entries/{txid}`, and then include the parameters as a separate field. (The CCF logging example does something like that, logging the method rather than the path)

app/src/tracing.h
letmaik: We can implement W3C trace context for the next release if needed but it's a bit more work than the two x-ms headers, therefore I didn't do it here. The main focus is on improving the logging for this PR and have some correlation identifiers.

app/src/main.cpp
letmaik: Sure, let's do that in the future sometime.

app/src/tracing.h
letmaik: Good point, ok, I think I'll add `make_endpoint` after all in this PR.

app/src/main.cpp
```diff
@@ -828,15 +820,14 @@ namespace scitt
...
-          error_adapter(ccf::json_adapter(get_version)), get_time),
+        ccf::json_adapter(get_version),
         no_authn_policy)
         .set_auto_schema<void, GetVersion::Out>()
         .set_forwarding_required(ccf::endpoints::ForwardingRequired::Never)
         .install();
 
 #ifdef ENABLE_PREFIX_TREE
-      // TODO add tracing_adapter() to all endpoints
+      // TODO local variant of make_endpoint in all endpoints
```
plietar, line 830: So I think this might "just work", in that the prefix tree code calls `registry.make_endpoint`, so it should pick up the new implementation.

On the other hand, it also calls `make_command_endpoint` and `make_read_only_endpoint`, which are what would need to be added here.

app/src/tracing.h
letmaik: I don't think a higher resolution is useful here, but using floating point seconds may be worth considering if that's more standard. I'll have a look.

letmaik: I think I prefer keeping it as-is. If we really need a higher resolution we can always switch to floating ms later on, but I doubt it's useful to get that kind of resolution.

plietar (approved):

PR #85, opened 2023-01-20
Include all past service identities in the DID endpoint.

### LINKED ISSUE #53: Define service DID document and add `did:web` endpoint

plietar:
This replaces the direct KV lookup with an indexing strategy that collects and keeps in memory all past identities. This allows the endpoint to return a full DID document with all these identities, without the need for historical queries and retries.

This PR adds support for disaster recovery to our test infrastructure, by restarting the cchost process. Because cchost port numbers are randomly assigned by the operating system, there would not be a stable address on which to contact the service. This is problematic both because we'd have to update `Client` instances to point to the new service, and because the service's address is used as its did:web long term identifier. To circumvent both problems, a Python TCP proxy is used, listening on a unique port for the entire duration of a test, and dynamically switching the destination of connections as the cchost process is restarted.

Fixes #53


test/infra/proxy.py
```diff
@@ -0,0 +1,98 @@
...
+
+import aiotools
+from loguru import logger as LOG
+
+from .async_utils import EventLoopThread, race_tasks
+
+
+class Proxy(EventLoopThread):
+    """
+    A TCP proxy, running as a context manager.
```
letmaik, line 15: What are we doing when HTTP/3 is a thing? That would be UDP based. Would we need to maintain two proxy implementations?

test/infra/proxy.py
```diff
@@ -0,0 +1,98 @@
...
+
+    The proxy listens for connections on a socket, and for each incoming
+    connection creates a new outgoing connection to a pre-configured upstream
+    host and port number. It then forwards data back and forth between the two
+    sockets.
+
+    The configured upstream can be modified after instantiation by calling
+    `set_upstream`. The change only affects future connections, not existing
```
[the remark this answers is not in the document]
plietar: If and when the upstream shuts down, it closes all its connections, and in-turn the proxy shuts down the corresponding incoming connection. So this works because in our use-case where we shut down the old service.

If we were to use this to switch between running services we might run into the issue you've described. I'm not sure it is worth fixing this now, although I'll add a note to explain this.

test/infra/async_utils.py
```diff
@@ -0,0 +1,162 @@
+# Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
plietar: Surprisingly enough, nothing I could find.

https://github.com/groove-x/trio-util has `wait_any` which is the same as `race_tasks`, but this is based around trio rather than asyncio. There's also a [lengthy discussion about it on the trio issue tracker](https://github.com/python-trio/trio/issues/1089).

I couldn't find any precedent for EventLoopThread.

test/infra/cchost.py
```diff
@@ -123,75 +109,45 @@ def __init__(
...
-        Entrypoint for the background thread.
+    async def run(self) -> None:
         """
-        try:
-            self.loop.run_until_complete(self._start_service())
-        except BaseException as e:
-            with self.cond:
-                self.thread_exception = e
-                self.cond.notify_all()
+        Invoked by EventLoopThread when the Proxy is started.
```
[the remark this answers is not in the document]
plietar: Yeah I'm not sure how this got here. Definitely meant something else

test/infra/proxy.py
plietar: Does CCF actually have any short/medium term plans on adding HTTP/3? But generally yes, we would probably need a second proxy.

plietar: This also assumes we even have an HTTP/3 client, which we currently don't.

letmaik: This is not a concern for now but I'm just trying see how much complexity having a proxy in the tests adds.

letmaik:
For backwards-compatibility we'll still have the `/parameters/historic` endpoint and currently this only includes the first service cert after a DR. I think we should expose the same data there as is included in the DID endpoint here and rely on the same indexer.

plietar:
Yeah good point. I think it makes sense to have them return equivalent data.

letmaik (approved):

PR #86, opened 2023-01-25
Do not treat app/src as a system headers search path.

plietar:
The `add_ccf_app` CMake function has a feature to add directories to the include search path of the compilation target. We were using that feature to add, among others, the main application folder `app/src`.

However, that macro adds these directories as "system directories", ie. CMake will use the `-isystem` compiler flag when building the command line argument. This in turn instructs Clang to silence any warning stemming from headers found in these paths.  See the [Clang user manual] for more details.

The end result was that our parts of our own code was being compiled with all warnings suppressed.

This commit fixes the CMake configuration, by manually calling `target_include_directories`, making sure thee paths we add are not treated as system paths. It also fixes a few warnings that had gone unnoticed until now.

In the future we should upstream a fix to CCF, allowing users of the `add_ccf_app` to specify both systems and regular search paths.

[Clang user manual]: https://releases.llvm.org/10.0.0/tools/clang/docs/UsersManual.html#controlling-diagnostics-in-system-headers


app/CMakeLists.txt
```diff
@@ -47,10 +47,20 @@ configure_file(src/generated/constants.h.in src/generated/constants.h @ONLY)
...
 )
 
+# The INCLUDE_DIRS argument of the add_ccf_app macro adds directories with
+# -isystem, which silences any warnings in them, which is undesirable for our
+# own code. We manually add them instead to workaround this and make sure -I
+# gets used.
+if (COMPILE_TARGET STREQUAL "sgx")
+  target_include_directories(scitt.enclave PRIVATE src ${CMAKE_CURRENT_BINARY_DIR}/src)
+elseif (COMPILE_TARGET STREQUAL "virtual")
+  target_include_directories(scitt.virtual PUBLIC src ${CMAKE_CURRENT_BINARY_DIR}/src)
```
letmaik, line 61:
```suggestion
  target_include_directories(scitt.virtual PRIVATE src ${CMAKE_CURRENT_BINARY_DIR}/src)
```

letmaik (approved):

PR #93, opened 2023-02-02
Don't hardcode localhost as resolver callback address

### LINKED ISSUE #92: Don't hardcode `localhost` in `did:web` resolver as callback address
https://github.com/microsoft/scitt-ccf-ledger/blob/28a7820f2944913fd5a2bd5d4d0536b6cee5daa4/app/src/did/web/method.h#L53-L55

If a node is configured to bind to `0.0.0.0` then using `localhost` works. But if it binds to a concrete IP address that's not `0.0.0.0`, then `localhost` doesn't work.

The code should be changed to just use the binding address as callback address as-is. Note that this works even with `0.0.0.0` (which is  mostly a Linux feature).

letmaik:
Fixes #92

### DIFF
--- a/app/src/did/web/method.h
+++ b/app/src/did/web/method.h
@@ -48,6 +48,8 @@ namespace scitt::did::web
         const auto& primary_interface =
           node_info->rpc_interfaces.begin()->second;
 
+        // Note that bind_address can be 0.0.0.0. This is fine here
+        // as Linux routes that address to localhost.
         return fmt::format(
           "https://{}/did/{}/doc", primary_interface.bind_address, did);
       }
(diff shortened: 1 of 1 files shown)


app/src/did/web/method.h
```diff
@@ -48,11 +48,8 @@ namespace scitt::did::web
         const auto& primary_interface =
           node_info->rpc_interfaces.begin()->second;
 
-        // We ignore the hostname part of the bind address, since that is
-        // typically something like 0.0.0.0, and hardcode localhost instead.
-        auto [_host, port] =
-          ccf::split_net_address(primary_interface.bind_address);
-        return fmt::format("https://localhost:{}/did/{}/doc", port, did);
+        return fmt::format(
```
plietar, line 51: I think this deserves a comment noting that `bind_adress` could be 0.0.0.0, but it works fine, given this behaviour is actually a little surprising.

plietar (approved):

app/src/did/web/method.h
letmaik: Done.

PR #97, opened 2023-02-06
Fix a few decoding bugs, and increase test coverage.

plietar:
- The content-type parameter was decoded as QCBOR_TYPE_ANY, but then
  assumed to always be a string. This could cause a type confusion and
  lead to crashes. It is now changed to be accept strings on integers
  only.

- The handling of a missing alg header parameter was creating an
  exception but not throwing it. This didn't weaken validation, since a
  bad_optional exception would be thrown two lines further down instead,
  but produced confusing and unexpected error responses.

- Added tests for various encoding and invalid messages edge cases. We
  cannot rely on pycose for many of these as it refuses malformed
  messages, so we use cbor2 directly.

- When embedding a receipt in a Sign1 enveloppe, which contained a
  single unprotected X509 certificate x5chain, the wrong QCBOR function
  was used, producing incorrectly formatted outputs.

- Check the return value when decoding X509 certificates and throw a
  ValidationError. Previously this could cause NULL pointer
  dereferencing.

### DIFF
--- a/test/test_ccf.py
+++ b/test/test_ccf.py
@@ -119,21 +119,6 @@ def test_consistent_kid(client, did_web, trust_store):
     assert did_doc["assertionMethod"][0]["id"] == f"{identity.issuer}{kid}"
 
 
-def test_invalid_kid(client, did_web):
-    """
-    Submit a claim with an invalid kid and check that it is rejected.
-    """
-    identity = did_web.create_identity(kid="#key-1")
-    invalid_identity = crypto.Signer(
-        identity.private_key, issuer=identity.issuer, kid="key-1"
-    )
-
-    claim = crypto.sign_json_claimset(invalid_identity, {"foo": "bar"})
-
-    with pytest.raises(ServiceError, match="kid must start with '#'"):
-        client.submit_claim(claim)
-
-
 @pytest.mark.needs_cchost
 @pytest.mark.isolated_test
 def test_recovery(client, did_web, restart_service):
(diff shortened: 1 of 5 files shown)


test/test_ccf.py
```diff
@@ -172,6 +190,32 @@ def test_submit_claim_notary_x509(
...
+        client.submit_claim(claims)
+
+
+def test_root_identity(
+    client: Client,
+    trust_store: TrustStore,
+    trusted_ca: X5ChainCertificateAuthority,
+):
+    """
+    Submit a claim signed by the self-signed root certificate.
```
letmaik, line 211: A certificate with the CA bit set should never be used for anything other than signing certificates that are issued. The fact that we allow this is a bug. In the trust store, there should only be CA certificates or self-signed non-CA certificates.

plietar:
Changes to X509 validation was broken off into PR #104.

letmaik:
@plietar X.509 changes are still in here, right?

plietar:
@letmaik it is based on the other PR, just because I've moved stuff a bit in the test infra, but it made Github display all of it. I've change the target branch so only the encoding fixes should be visible now.

letmaik (approved):
Should be merged into main once the other PR is in.

plietar:
@letmaik made some minor new changes following the rebase on the x509 PR. Main one being that I realized the service would crash on a ill-formed DER in the x5chain. Fixed now.

letmaik (approved):

PR #99, opened 2023-02-10
Implement the new REST API

### LINKED ISSUE #108: Implement updated REST API
See https://github.com/ietf-wg-scitt/draft-ietf-scitt-architecture/pull/3.

plietar:
This introduces the concept of a long-running operation to the API. When
submitting a claim, the ledger returns a temporary operation ID. Clients
must poll an endpoint to check the status of the operation.

Eventually, the operation will complete and the status endpoint will
return long-term entry ID, which the client can use to retrieve a
receipt.

This schema can be used to model both the DID-web resolution phase and
the CCF consensus process.

If submitting a claim requires a DID resolution, the initial POST to the
`/claims` endpoint creates a new operation by saving the claim to a
special KV, and triggers the external process to fetch the DID document.
When the external process invokes the operation callback with the
resolution payload, the ledger uses a historical to get the original
claim back. Once both the claim and DID document are available, the
existing registration process takes place and a new row is added to the
KV to mark the operation as complete.

After a claim has been written the the KV, either synchronously in the
original POST request or asynchronously in the resolution callback, the
operation is still considered as pending, until the transaction is
globally committed and a receipt is available. At this point only will
the operation be considered completed.

This new API has multiple benefits:
- It surfaces DID resolution errors to the end-user. Previously, clients
  would only ever receive "resolution in progress" messages, but never
  any hard errors with informative description. Now the operation can be
  marked as failed by writing details of the error to the KV, and the
  error message is returned when a client polls the operation status.
- It is a more intuitive API and does not require clients to retry POST
  requests anymore. The only request that needs polling is a GET to
  check the operation status, and one day this may even be replaced by a
  long-polling request.
- It unifies different delays, such as resolution and consensus, under a
  single interface, abstracting implementation details from clients.

Fixes #108

### DIFF
--- a/app/src/call_types.h
+++ b/app/src/call_types.h
@@ -2,11 +2,6 @@
 // Licensed under the MIT License.
 
 #pragma once
-#ifdef VIRTUAL_ENCLAVE
-#  include "did/unattested.h"
-#else
-#  include "did/attested.h"
-#endif
 #include "kv_types.h"
 
 #include <ccf/ds/json.h>
@@ -16,15 +11,6 @@
 
 namespace scitt
 {
-  struct PostDidResolution
-  {
-#ifdef VIRTUAL_ENCLAVE
-    using In = did::UnattestedResolution;
-#else
-    using In = did::AttestedResolution;
-#endif
-  };
-
   struct GetIssuerInfo
   {
     using Out = IssuerInfo;
@@ -100,5 +86,43 @@ namespace scitt
 
   DECLARE_JSON_TYPE(GetVersion::Out);
   DECLARE_JSON_REQUIRED_FIELDS(GetVersion::Out, scitt_version);
+
+  struct GetEntry
+  {
+    struct Out
+    {
+      ccf::TxID entry_id;
+    };
+  };
+
+  DECLARE_JSON_TYPE(GetEntry::Out);
+  DECLARE_JSON_REQUIRED_FIELDS_WITH_RENAMES(GetEntry::Out, entry_id, "entryId");
+
+  struct GetOperation
+  {
+    struct Out
+    {
+      ccf::TxID operation_id;
+      OperationStatus status;
+      std::optional<ccf::TxID> entry_id;
+      std::optional<nlohmann::json> error;
+    };
+  };
+
+  DECLARE_JSON_TYPE_WITH_OPTIONAL_FIELDS(GetOperation::Out);
+  DECLARE_JSON_REQUIRED_FIELDS_WITH_RENAMES(
+    GetOperation::Out, operation_id, "operationId", status, "status");
+  DECLARE_JSON_OPTIONAL_FIELDS_WITH_RENAMES(
+    GetOperation::Out, entry_id, "entryId", error, "error");
+
+  struct GetAllOperations
+  {
+    struct Out
+    {
+      std::vector<GetOperation::Out> operations;
+    };
+  };
+  DECLARE_JSON_TYPE(GetAllOperations::Out);
+  DECLARE_JSON_REQUIRED_FIELDS(GetAllOperations::Out, operations);
 
 } // namespace scitt
--- a/app/src/operations_endpoints.h
+++ b/app/src/operations_endpoints.h
@@ -575,7 +575,7 @@ namespace scitt
    *
    * Asynchronous operations are useful for tasks which depend on an external
    * process to execute. This transaction will record the start of the operation
-   * and set its status to "pending". When the external process completes, it
+   * and set its status to "running". When the external process completes, it
    * must invoke a callback URL, at which point more processing will happen
    * inside the ledger.
    *
--- a/app/src/operations_endpoints.h
+++ b/app/src/operations_endpoints.h
@@ -1,0 +1,681 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include "app_data.h"
+#include "historical/historical_queries_adapter.h"
+#include "odata_error.h"
+#include "visit_each_entry_in_value.h"
+
+#include <ccf/json_handler.h>
+#include <ccf/node_context.h>
+#include <ccf/service/tables/nodes.h>
+
+namespace scitt
+{
+  using OperationCallback = std::function<void(
+    ccf::endpoints::EndpointContext&, const nlohmann::json&, nlohmann::json&&)>;
+
+  /**
+   * An indexing strategy which maintains a map from Operation ID to state.
+   *
+   * The state can be one of "running", "failed" or "succeeded". In the last
+   * case, the transaction ID which completed the operation is recorded.
+   */
+  class OperationsIndexingStrategy
+    : public VisitEachEntryInValueTyped<OperationsTable>
+  {
+  public:
+    OperationsIndexingStrategy(ccf::BaseEndpointRegistry& registry) :
+      OperationsIndexingStrategy(
+        [&registry](timespec& time) {
+          return registry.get_untrusted_host_time_v1(time);
+        },
+        [&registry](
+          ccf::View view, ccf::SeqNo seqno, ccf::TxStatus& tx_status) {
+          return registry.get_status_for_txid_v1(view, seqno, tx_status);
+        })
+    {}
+
+    OperationsIndexingStrategy(
+      std::function<ccf::ApiResult(timespec& time)> get_time,
+      std::function<ccf::ApiResult(
+        ccf::View view, ccf::SeqNo seqno, ccf::TxStatus& tx_status)>
+        get_status_for_txid) :
+      VisitEachEntryInValueTyped(OPERATIONS_TABLE),
+      get_time(get_time),
+      get_status_for_txid(get_status_for_txid)
+    {}
+
+    /**
+     * Get the list of all operations whose state is known by the indexing
+     * strategy.
+     *
+     * The function isn't intended to be used in normal situations, but can
+     * serve as a convenient debugging endpoint into the state of the indexing
+     * strategy.
+     */
+    std::vector<GetOperation::Out> operations() const
+    {
+      std::lock_guard guard(lock);
+      std::vector<GetOperation::Out> result;
+      for (const auto& it : operations_)
+      {
+        result.push_back(GetOperation::Out{
+          .operation_id = ccf::TxID{it.second.view, it.first},
+          .status = it.second.status,
+          .entry_id = it.second.completion_tx,
+          .error = it.second.error,
+        });
+      }
+      return result;
+    }
+
+    /**
+     * Look up an operation by its transaction ID.
+     *
+     * If the transaction ID is unknown, this function generally returns an
+     * operation in the "running" state anyway, since it would not be possible
+     * to distinguish between whether this is an invalid transaction ID or if
+     * the transaction that created the operation has not been globally
+     * committed and indexed yet.
+     *
+     * In some cases however, we can be confident that this transaction ID is
+     * invalid now and forever in the future. In these cases, an appropriate
+     * HTTPError exception is thrown.
+     */
+    GetOperation::Out lookup(const ccf::TxID& operation_id) const
+    {
+      std::lock_guard guard(lock);
+
+      ccf::TxStatus tx_status;
+      auto result =
+        get_status_for_txid(operation_id.view, operation_id.seqno, tx_status);
+      if (result != ccf::ApiResult::OK)
+      {
+        throw InternalError(fmt::format(
+          "Failed to get transaction status: {}",
+          ccf::api_result_to_str(result)));
+      }
+
+      switch (tx_status)
+      {
+        case ccf::TxStatus::Unknown:
+        case ccf::TxStatus::Pending:
+          return {
+            .operation_id = operation_id,
+            .status = OperationStatus::Running,
+          };
+
+        case ccf::TxStatus::Invalid:
+          // This state can arise even in a well-behaved client if the view
+          // changed (eg. because of a Raft election), and the operation's
+          // transaction got dropped. It could also be a client giving us
+          // garbage txids, but we can't tell the difference so we remain polite
+          // and assume the former and say the operation has failed.
+          return {
+            .operation_id = operation_id,
+            .status = OperationStatus::Failed,
+            .error =
+              ODataError{
+                .code = ccf::errors::TransactionInvalid,
+                .message = "Transaction is invalid",
+              },
+          };
+
+        case ccf::TxStatus::Committed:
+          // This is the main case, when the client is referring to a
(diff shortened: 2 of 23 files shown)


api.md
```diff
@@ -0,0 +1,185 @@
...
+One of the following
+- Status 202 - Registration is pending
+    - Header: `Location: <Base URL>/operations/<Operation ID>`
+    - Header: `Content-Type: application/json`
+    - Body: `{ "operationId": "<Operation ID>", "status": "pending" }`
+
+- Status 400 - Registration was unsuccessful due to invalid input
+    - Error code `InvalidInput`
+
+- Status 201 - Registration was successful
```
letmaik, line 18: I think it's better to have 202 always and if it's completed immediately, then the status is `"registered"`.

api.md
```diff
@@ -0,0 +1,185 @@
...
+# Retrieve claim
+
+## Request
+
+`GET <Base URL>/entries/<Entry ID>/claim`
+
+## Response
+
+- Status 200
+  - Header `Content-Type: application/cose`
```
letmaik, line 73: Add Body

api.md
```diff
@@ -0,0 +1,185 @@
...
+# Retrieve receipt
+
+## Request
+
+`GET <Base URL>/entries/<Entry ID>/receipt`
+
+## Response
+
+- Status 200
+  - Header `Content-Type: application/cbor`
```
letmaik, line 84: Add body

api.md
```diff
@@ -0,0 +1,185 @@
...
+
+# CCF based design
+
+## Operations table
+
+This is a singleton table, without keys. Each record can be one of the following:
+
+- `{ "status": "registered" }`: A claim was registered synchronously. The Operation ID and Entry ID are both equal to the transaction ID of this record.
+- `{ "status": "pending", "startTime": TIME, "contextDigest": "..." }`: A claim was submitted and is being processed asynchronously. The Operation ID is equal to the transaction ID of this record. The entry ID is not yet know.
```
letmaik, line 93:
```suggestion
- `{ "status": "pending", "startTime": TIME, "contextDigest": "..." }`: A claim was submitted and is being processed asynchronously. The Operation ID is equal to the transaction ID of this record. The entry ID is not yet known.
```

api.md
```diff
@@ -0,0 +1,185 @@
...
+    - Sets the claims digest of the transaction to the receipt TBS.
+3) The local commit handler takes the transaction number and returns it to the
+    client in a JSON document as the operation ID and `"status": "pending"`.
+4) Client requests `GET /operations/<Operation ID>`
+    - The indexing strategy is unaware of the operation ID
+    - The endpoint returns `"status": "pending"`
+5) CCF replicates the transaction and its signature.
+6) The indexing strategy witnesses the write to the `operations` table and
+   updates its map, using the transaction ID as both the Operation and Entry ID.
+7) Client requests `GET /operations/<Operation ID>`. Using the indexing
```
letmaik, line 133: I'm wondering if there may still be an issue with how global commits work: https://microsoft.github.io/CCF/main/build_apps/kv/kv_how_to.html#global-commit. A global commit may be observed by a given node when then returns `registered` but a follow-up GET to another node that's a little behind may not have the current state yet.

api.md
```diff
@@ -0,0 +1,185 @@
...
+   strategy, the endpoint returns `"status": "registered"` and the appropriate
+   Entry ID.
+8) Client requests `GET /entries/<Entry ID>/receipt`. It may receiver a transient
```
letmaik, line 136:
```suggestion
8) Client requests `GET /entries/<Entry ID>/receipt`. It may receive a transient
```

api.md
```diff
@@ -0,0 +1,185 @@
...
+     and performs full validation of the claim.
+   - The ledger writes the claim to the `entry` table, and
+     `{ "operationId": "<Operation ID>", "status": "registered" }` to the
+     operations table.
+9) The indexing strategy witnesses the second write to the `operations` table
+   and updates its map, using the transaction ID as the Entry ID, setting the
+   status to registered.
+10) Client requests `GET /operations/<Operation ID>`. Using the indexing
+   strategy, the endpoint returns `"status": "registered"` and the appropriate
+   Entry ID.
```
letmaik, line 178: Should have the final step of fetching the receipt like in the sync flow.

letmaik (commented):
Had an initial look at the docs, I was wondering about global commit semantics. See comments.

app/src/operations_endpoints.h
```diff
@@ -0,0 +1,402 @@
...
+          errors::IndexingInProgressRetryLater,
+          fmt::format("Historical data is not ready yet (seq: {})", id.seqno));
+      }
+
+      auto htx = hstate->store->create_read_only_tx();
+      auto historical_operation =
+        htx.ro<OperationsTable>(OPERATIONS_TABLE)->get();
+      if (!historical_operation.has_value())
+      {
+        throw BadRequestError(errors::InvalidInput, "FOO");
```
letmaik, line 187: Could be a 404 even.

app/src/operations_endpoints.h
```diff
@@ -0,0 +1,402 @@
...
+        HTTP_POST,
+        ccf::json_adapter(std::bind(
+          endpoints::post_operation_callback, callback, &state_cache, _1, _2)),
+        no_authn_policy)
+      .install();
+  }
+
+  /**
+   * Get the address and port number on which the RPC interface is listening.
+   * This is forms the base of the callback URL used by the external process to
```
letmaik, line 268:
```suggestion
   * This forms the base of the callback URL used by the external process to
```

app/src/operations_endpoints.h
```diff
@@ -0,0 +1,402 @@
...
+      .context = std::move(callback_context),
+    });
+
+    // The AppData allows us to propagate data to the handler. We use it to keep
+    // a reference to the trigger lambda and the bind address. The latter is
+    // needed to build the callback URL, but needs read access to the KV, which
+    // we won't have later on in the commit handler.
+    get_app_data(endpoint_context.rpc_ctx).asynchronous_operation =
+      AsynchronousOperation{
+        .trigger = trigger,
```
letmaik, line 336: This is definitely not something we'd be able to do with CCF's external executors, at most they would allow some JSON type user data.

app/src/operations_endpoints.h
```diff
@@ -0,0 +1,402 @@
...
+    // only pending.
+    GetOperation::Out operation;
+    operation.status = OperationStatus::Pending;
+    operation.operation_id = tx_id;
+
+    ctx.rpc_ctx->set_response_header(http::headers::CCF_TX_ID, tx_str);
+    ctx.rpc_ctx->set_response_header(
+      http::headers::CONTENT_TYPE, http::headervalues::contenttype::JSON);
+    ctx.rpc_ctx->set_response_header(
+      http::headers::LOCATION, fmt::format("/operations/{}", tx_str));
```
letmaik, line 389: Location must be an absolute URL.

pyscitt/pyscitt/client.py
```diff
@@ -364,34 +365,47 @@ def get_did_document(self, did: str) -> dict:
...
         if skip_confirmation:
-            return Submission(tx)
+            return PendingSubmission(operation_id)
         else:
+            tx = self.resolve_operation(operation_id)
             receipt = self.get_receipt(tx, decode=False)
-            return SubmissionReceipt(tx, receipt)
+            return Submission(operation_id, tx, receipt)
+
+    def resolve_operation(self, operation: str) -> str:
```
letmaik, line 395: Maybe `wait_for_operation`?

app/src/operations_endpoints.h
plietar: Does it? https://www.rfc-editor.org/rfc/rfc9110.html#section-10.2.2 seems to say relative is fine.

app/src/operations_endpoints.h
```diff
@@ -0,0 +1,402 @@
...
+    {
+      OperationStatus status;
+      std::optional<ccf::TxID> completion_tx;
+      std::optional<nlohmann::json> error;
+    };
+
+    std::optional<OperationState> lookup(const ccf::TxID& operation_id) const
+    {
+      std::lock_guard guard(lock);
+      auto it = operations.find(operation_id.seqno);
```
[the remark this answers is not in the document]
plietar: `ccf::TxID` doesn't have a defined comparison operator (probably for good reasons), which means it can't be used in an `std::map`.

Keys in the `std::map` always refer to committed transactions, so their sequence number is guaranteed to be unique. I guess there should be a check on lookup though to make sure the view is correct.

app/src/operations_endpoints.h
plietar: Yeah that's probably right.

I think I'd rather keep it as is for now. Using a lambda here is nice, since it means the operations code can be very agnostic of how they are actually implemented, and doesn't need to explicitly capture references to CCF state (eg. the AbstractNodeContext needed to get the AbstractHostProcesses object).

app/src/operations_endpoints.h
letmaik: Huh, I wasn't aware of that. Looks like the original HTTP 1.1 RFC only allows absolute but the replacement RFC allows both. I haven't seen a relative URL in APIs myself and given that the new RFC just came out last year I'm thinking we should still use absolute URLs for now. We can construct the full URL using the hostname like here: https://github.com/microsoft/CCF/blob/9f7534f0d9af4a95a3ebeb09a1ad62eeed680c22/src/apps/js_generic/js_generic_base.cpp#L189-L200

app/src/call_types.h
```diff
@@ -101,4 +87,20 @@ namespace scitt
   DECLARE_JSON_TYPE(GetVersion::Out);
   DECLARE_JSON_REQUIRED_FIELDS(GetVersion::Out, scitt_version);
 
+  struct GetOperation
+  {
+    struct Out
+    {
+      OperationStatus status;
+      ccf::TxID operation_id;
```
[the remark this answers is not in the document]
plietar: Done

app/src/operations_endpoints.h
plietar: Based on `https://github.com/microsoft/CCF/issues/4996`, this seems fine actually.

If anything, it looks like we won't even need the AppData anymore and the two phases of the request will happen in the same context.

app/src/operations_endpoints.h
plietar: The similar `/entries/<txid>` uses a 400 so I've used the same to stay consistent. Fixed the error message though,

app/src/operations_endpoints.h
```diff
@@ -0,0 +1,681 @@
...
+
+        case ccf::TxStatus::Committed:
+          // This is the main case, when the client is referring to a
+          // transaction that has been committed to the ledger. From now on, it
+          // is safe to consider the SeqNo only and compare it to what has been
+          // indexed.
+          break;
+      }
+
+      if (operation_id.seqno < lower_bound)
```
letmaik, line 133: This would also be returned when the out-of-range transaction wasn't an operation transaction. I think to keep it simple it's better to just return 404 like below.

app/src/operations_endpoints.h
```diff
@@ -0,0 +1,681 @@
...
+          ccf::api_result_to_str(result)));
+      }
+
+      switch (tx_status)
+      {
+        case ccf::TxStatus::Unknown:
+        case ccf::TxStatus::Pending:
+          return {
+            .operation_id = operation_id,
+            .status = OperationStatus::Running,
```
letmaik, line 106: Note to self: if we stick with "running", update spec and emulator which currently use "pending".

app/src/operations_endpoints.h
```diff
@@ -0,0 +1,681 @@
...
+      throw std::runtime_error("Could not callback URL");
+    }
+  }
+
+  /**
+   * Mark the start of an asynchronous operation.
+   *
+   * Asynchronous operations are useful for tasks which depend on an external
+   * process to execute. This transaction will record the start of the operation
+   * and set its status to "pending". When the external process completes, it
```
letmaik, line 576: Note here it's still called `pending`.

pyscitt/pyscitt/client.py
```diff
@@ -313,29 +313,31 @@ def get_historical(self, *args, retry_on=[], **kwargs):
             retry_on=[
                 HTTPStatus.ACCEPTED,
                 (HTTPStatus.NOT_FOUND, "TransactionPendingOrUnknown"),
```
letmaik, line 315: Do we still need these two cases?

test/test_resolution.py
```diff
@@ -14,6 +15,36 @@
 from .infra.did_web_server import DIDWebServer
 
 
+@pytest.fixture
+def submit_concurrently(
+    client: Client,
+    did_web: DIDWebServer,
+):
+    """
+    A fixture which allows multiple receipts to be submitted concurrently.
```
letmaik, line 24:
```suggestion
    A fixture which allows multiple claims to be submitted concurrently.
```

test/test_resolution.py
```diff
@@ -135,6 +223,107 @@ def test_claim_without_kid(
...
+    """
+
+    @pytest.fixture
+    def identity(self, did_web) -> crypto.Signer:
+        issuer = did_web.generate_identifier()
+        private_key, _ = crypto.generate_keypair(kty="ec")
+        return crypto.Signer(private_key, issuer=issuer)
+
+    @pytest.fixture(autouse=True)
+    def create_document(self, did_web: DIDWebServer, identity: crypto.Signer):
```
letmaik, line 243: I find this instance of `autouse` a bit confusing when looking at the test functions. Maybe making it more explicit would make it easier to understand the tests?

app/src/operations_endpoints.h
plietar: This code generally assumes the client is well-behaved and is only using actual operations, so it gives it the benefit of the doubt and replies accordingly. This is why this returns a 410 rather than 404.

Only when we know for certain this transaction couldn't have been an operation do we return a 404.

app/src/operations_endpoints.h
letmaik: Done. Renamed to "running".

app/src/operations_endpoints.h
plietar: Done.

pyscitt/pyscitt/client.py
plietar: The prefix tree code was still using Accepted for this. I've switched it to ServiceUnavailable to be consistent and removed these.

test/test_resolution.py
letmaik: That's better, thanks.

letmaik (approved):

PR #103, opened 2023-02-13
Config documentation

darracott:
Adds a page documenting SCITT config options. 

Addresses https://github.com/microsoft/scitt-ccf-ledger/issues/32

PR #104, opened 2023-02-13
Make x509 certificate validation stricter.

plietar:
Certificate chains embedded in claims (as part of the x5chain parameter) now need to include the self-signed root certificate. This was previously not necessary, since the root certificate would be resolved from the service's trust store. Requiring it improves auditability however, since it ensures the claim can be verified in isolation after the fact. This is also in line with Notary's policy.

Additionally, leaf certificates may not have the CA bit set in their basic constraints anymore.

Test coverage for certificate validation has been moved to its own file and broadly extended to cover many more cases.

This was split off of PR #97

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -129,6 +129,7 @@ if (BUILD_TESTS)
 
   add_executable(
     unit_tests
+    unit-tests/main.cpp
     unit-tests/cbor_test.cpp
     unit-tests/verifier_test.cpp
     unit-tests/configurable_auth_test.cpp
@@ -142,7 +143,7 @@ if (BUILD_TESTS)
   target_compile_options(unit_tests PUBLIC ${COMPILE_LIBCXX})
   target_link_libraries(
     unit_tests PRIVATE
-    GTest::gmock_main
+    GTest::gmock
     rapidcheck
     rapidcheck_gtest
     t_cose.host
--- a/app/src/verifier.h
+++ b/app/src/verifier.h
@@ -415,13 +415,18 @@ namespace scitt::verifier
       }
 
       OpenSSL::Unique_X509_STORE_CTX store_ctx;
-      OpenSSL::CHECK1(X509_STORE_set_flags(store, X509_V_FLAG_PARTIAL_CHAIN));
       OpenSSL::CHECK1(X509_STORE_CTX_init(store_ctx, store, leaf, chain_stack));
 
       if (X509_verify_cert(store_ctx) != 1)
       {
+        int err = X509_STORE_CTX_get_error(store_ctx);
+        SCITT_INFO(
+          "Certificate chain is invalid: {}",
+          X509_verify_cert_error_string(err));
         throw VerificationError("Certificate chain is invalid");
       }
+
+      check_certificate_policy(chain_stack, leaf);
 
       return leaf;
     }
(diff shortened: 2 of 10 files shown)


plietar:
I've been getting inconsistent results in CI for this PR, with the tests passings on SGX but not with the virtual build. I think I've narrowed it down to a difference in OpenSSL version: OpenEnclave uses OpenSSL 1.1.1q whereas the Ubuntu-based image we use only ships OpenSSL 1.1.1f.

The older OpenSSL rejects self-signed certificates which don't have the CA flag set, but this is accepted on the newer builds. This seems to be the upstream PR that introduced the change: https://github.com/openssl/openssl/pull/12357, and the original issue https://github.com/openssl/openssl/issues/1418.

In short, it seems we are trying to satisfy the following 3 requirements, which leads to a contradiction, and we have to drop one.
- Require self-signed trust anchor certificates to have the CA flag set (which old OpenSSL forces on us)
- Reject end-entity certificates that have the CA flag set (which this PR was trying to do)
- Support self-signed end-entity certificates, provided they are in the trust store

I'm leaning towards dropping the latter, since IMO self-signed EE certificates have limited benefits in practical deployments, even though they can be handy for simple demos.

If we do in fact decide to reject self-signed end-entity certs then we should do so in a consistent way, irrespective of the OpenSSL version. From my reading of the OpenSSL PR linked above, this can be achieved by setting the X509_V_FLAG_CHECK_SS_SIGNATURE flag, which even for recent versions means "check the self-signature, including any key usage"

(As a side note, if we reject self-signed EE certs, then we won't ever have to deal with x5chain parameters of length 1, which closes that can of worms)

----

There's actually one bit left which is slightly confusing: the unit tests include a case of verifying a self-signed EE certificate, without the CA flag set, and this works fine even on the older OpenSSL, where the same test fails as a integration test.

I think this comes down to the way the certificate is generated, and the fact the "CA" or "not CA" comes in different forms:
- The unit tests add a `CA:TRUE` or `CA:FALSE`, non-critical `BasicConstraint` extension to certificates, and does not add any `KeyUsage` extension.
- The integration tests adds a `CA:TRUE` or `CA:FALSE` *critical* `BasicConstraint`, as well as a *critical* `KeyUsage` where `KeyCertSign` is set to the same true or false value.

I don't think the criticality is too relevant here, but the KU extension is. The old OpenSSL versions only use this extension when rejecting self-signed EE certificates.

This opens up a "loophole": we could accept self-signed EE certificates, but only if the the KeyUsage extension is absent.

plietar:
Of course another option would be to upgrade our OpenSSL version used by virtual builds, although our hands are kind of tied by Ubuntu and CCF. Interestingly, I found a reference to this issue in the Ubuntu bug tracker: https://bugs.launchpad.net/ubuntu/+source/openssl/+bug/1875781.

letmaik:
I think upgrading OpenSSL is not realistic as that would mean asking CCF to use a different Docker base image just for the virtual build. As soon as Open Enclave supports 22.04, then CCF will upgrade to that, but I don't see an issue tracking that in OE so I don't expect it soon.

Accepting self-signed EE certs with non-critical key usage would go against the Notary spec which requires that to be present and critical, so that's not a great option either.

In the context of SCITT I think I agree that supporting self-signed EE certs doesn't make much sense, so I'm happy to remove that option even though it means not fully supporting Notary. This limitation should be mentioned in #103 (saying that trust anchors must not be self-signed EE certs). If there is demand for that feature in the future, we can always add support for it, maybe then more easily via an update of OpenSSL.

letmaik (approved):

Issue #105, opened 2023-02-16
Extend attested-fetch to return errors


letmaik:
https://github.com/microsoft/attested-fetch

Currently, any network error including HTTP errors (>=400) don't generate a result when calling attested-fetch. The error is dumped to console and the exit code is 1. It would be useful, especially for HTTP errors, to return an attested result that includes the error. This can then be surfaced in `GET /operations/<id>` to provide more details when DID resolution failed.

letmaik:
I created https://github.com/microsoft/attested-fetch/issues/6 for this.

PR #106, opened 2023-02-16
Rewrite did:web fetch scripts in Python

letmaik:
This has better logging and also adds retry logic.

### DIFF
--- a/app/fetch-did-web-doc.py
+++ b/app/fetch-did-web-doc.py
@@ -8,6 +8,7 @@ import ssl
 import subprocess
 import tempfile
 import time
+import http
 from urllib.error import HTTPError, URLError
 from urllib.request import Request, urlopen
 
@@ -100,7 +101,7 @@ def request(url, data=None, headers=None):
             body = e.read()
             retry_after = int(e.headers.get("Retry-After", HTTP_DEFAULT_RETRY_AFTER))
             logging.error(f"HTTP status {status_code}: {body}")
-            if status_code not in [429, 503]:
+            if status_code not in [http.HTTPStatus.TOO_MANY_REQUESTS, http.HTTPStatus.SERVICE_UNAVAILABLE]:
                 break
         except URLError as e:
             retry_after = HTTP_DEFAULT_RETRY_AFTER
--- a/app/fetch-did-web-doc.py
+++ b/app/fetch-did-web-doc.py
@@ -2,13 +2,13 @@
 # Licensed under the MIT License.
 import argparse
 import base64
+import http
 import json
 import logging
 import ssl
 import subprocess
 import tempfile
 import time
-import http
 from urllib.error import HTTPError, URLError
 from urllib.request import Request, urlopen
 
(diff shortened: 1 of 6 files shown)


app/fetch-did-web-doc.py
```diff
@@ -0,0 +1,142 @@
...
+            except subprocess.CalledProcessError as e:
+                logging.error(f"afetch failed with code {e.returncode}: {e.stdout}")
+            except Exception as e:
+                logging.error(f"Unknown error: {e}")
+            if retries >= 0:
+                logging.info(f"Retrying in {HTTP_DEFAULT_RETRY_AFTER} seconds")
+                time.sleep(HTTP_DEFAULT_RETRY_AFTER)
+            else:
+                raise e
+        if success:
```
plietar, line 65: Can this line ever be reached with `success = False`? It seems to me not, since when we always throw an exception when we run out of retries.

Then you don't even need the `success` variable

app/fetch-did-web-doc.py
```diff
@@ -0,0 +1,142 @@
...
+    with tempfile.NamedTemporaryFile() as out_path:
+        args = [
+            f"{AFETCH_DIR}/afetch",
+            f"{AFETCH_DIR}/libafetch.enclave.so.signed",
+            out_path.name,
+            url,
+            nonce,
+        ]
+        logging.info(f"Starting {' '.join(args)}")
+        while not success and retries >= 0:
```
plietar, line 48: I think this can be `while True` and keep the same behaviour.

app/fetch-did-web-doc.py
```diff
@@ -0,0 +1,145 @@
...
+            body = response.read()
+            success = True
+            logging.info(f"HTTP status {status_code}")
+            break
+        except HTTPError as e:
+            status_code = e.getcode()
+            body = e.read()
+            retry_after = int(e.headers.get("Retry-After", HTTP_DEFAULT_RETRY_AFTER))
+            logging.error(f"HTTP status {status_code}: {body}")
+            if status_code not in [429, 503]:
```
plietar, line 103: Replace these with constants from [`http.HTTPStatus`](https://docs.python.org/3/library/http.html#http.HTTPStatus)

app/fetch-did-web-doc.py
```diff
@@ -0,0 +1,145 @@
...
+                break
+        except URLError as e:
```
plietar, line 105: Should we be recording those errors and returning them to the ledger? Stuff like unresolvable hostnames or unable to establish a connection seems useful.

letmaik: I think recording that there was a network error is fine, though I think using the special status code 0 for that is enough. Note that this piece of code is only for the unattested case, and we'll do the same as part of #105 for attested. I think going from no error reporting to HTTP errors and whether there was a network error is good enough for now.

app/fetch-did-web-doc.py
letmaik: Good point, once attested-fetch supports returning attested errors then probably we want to have similar HTTP status based retry logic as for the unattested case.

plietar (approved):

Issue #101, opened 2023-02-13
Remove t_cose work-arounds for crit parameter


letmaik:
https://github.com/microsoft/scitt-ccf-ledger/pull/73 introduced some work-arounds that can be removed with t_cose 1.1.1. 

Depends on #100.

Issue #100, opened 2023-02-13
Update to t_cose 1.1.1


letmaik:
This will allow to remove some of the recently added work-around to support Notary header parameters.

This has to be happen in upstream CCF.

darracott:
https://github.com/microsoft/CCF/pull/5016#event-8527660244 updates `t_cose` to v.1.1.1 on the CCF 3.x support branch.  We'll now wait for a new 3.x release. Then we can pull it into SCITT and resolve https://github.com/microsoft/scitt-ccf-ledger/issues/101

achamayou:
@darracott https://github.com/microsoft/CCF/releases/tag/ccf-3.0.6

darracott:
Speedy! Thanks for pointing that out

darracott:
https://github.com/microsoft/scitt-ccf-ledger/pull/122

PR #107, opened 2023-02-17
Remove t_cose work-arounds for Notary style claims

darracott:
- [ISSUE-100](https://github.com/microsoft/scitt-ccf-ledger/issues/100): Update CCF from [3.0.2 to 3.0.6](https://github.com/microsoft/CCF/compare/ccf-3.0.2...ccf-3.0.6) which means a `t_cose` upgrade from [v1.1 to v1.1.1](https://github.com/laurencelundblade/t_cose/compare/v1.1...v1.1.1).
- [ISSUE-101](https://github.com/microsoft/scitt-ccf-ledger/issues/101): Replaced workaround code with `t_cose` for signature verification of Notary style claims.

Note: for testing, now that SCITT doesn't accept self-signed leaf certificates (which is what notation generates for testing), it's much easier to make use of X5ChainCertificateAuthority.create_chain() in test/infra/x5chain_certificate_authority.py to generate a cert chain with the correct extensions and flags, than to try to get the openssl incantation right.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Dependencies
+- [ISSUE-100](https://github.com/microsoft/scitt-ccf-ledger/issues/100): Update CCF from [3.0.2 to 3.0.6](https://github.com/microsoft/CCF/compare/ccf-3.0.2...ccf-3.0.6) which means a `t_cose` upgrade from [v1.1 to v1.1.1](https://github.com/laurencelundblade/t_cose/compare/v1.1...v1.1.1).
+
 ## [0.2.1]
 ### Changed
 - Include all past service identities in the DID endpoint (#85).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Fixed
+- [ISSUE-101](https://github.com/microsoft/scitt-ccf-ledger/issues/101): Replaced workaround code with `t_cose` for signature verification of Notary style claims.
+
 ### Dependencies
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+- [ISSUE-101](https://github.com/microsoft/scitt-ccf-ledger/issues/101): Replaced workaround code with `t_cose` for signature verification of Notary style claims.
```
letmaik, line 11: Please stick to existing format of changelog entries. When `#xy` is copied to a release entry it is automatically linked.

app/src/cose.h
```diff
@@ -691,7 +498,7 @@ namespace scitt::cose
     // auxiliary buffer size.
     t_cose_parameters params;
     t_cose_sign1_verify_init(
-      &verify_ctx, T_COSE_OPT_TAG_REQUIRED | T_COSE_OPT_DECODE_ONLY);
+      &verify_ctx, T_COSE_OPT_TAG_REQUIRED | T_COSE_OPT_DECODE_ONLY | T_COSE_OPT_UNKNOWN_CRIT_ALLOWED);
```
letmaik: Given that this also affects non-Notary verification I think the `process_*_profile` functions in `verifier.h` should be extended to check that the crit field is empty for the other profiles. And there should be a comment here saying why using this flag is ok.

plietar: Alternatively, add a flags argument to `cose::verify` that only the notary code path needs to set to `T_COSE_OPT_UNKNOWN_CRIT_ALLOWED`.

plietar:
So https://github.com/microsoft/CCF/pull/5000 seems to have broken the unit tests.

We were using CCF's api to generate the certs used in unit tests, but they now set a path length constraint which breaks anything using intermediate CAs. I think the only workaround is for us to copy CCF's certificate generation code into `verifier_test.cpp` (or remove those tests).

letmaik:
Can we do this in two steps? Updating of CCF should be a separate PR and we don't want to make that dependent on other work that's not strictly required for the update.

darracott:
Superseded by https://github.com/microsoft/scitt-ccf-ledger/pull/132

PR #110, opened 2023-02-20
Add a flag in cchost.py to run with GDB.

plietar:
The cchost infra (and start.sh) can now be given a `--with-gdb` argument, which will start the cchost process in a debugger. This can be a handy way to get stack traces when debugging crashes or exceptions.

While pytest cannot start a cchost process inside of GDB, most tests can still be used while connecting to an external service (as used in our Docker CI). This allows pytest to be using together with `./start.sh --with-gdb`, if trying to run tests under a debugger.

For now, using GDB probably only works on virtual builds. It might require a bit more setup to function on SGX builds.

### DIFF
--- a/test/infra/async_utils.py
+++ b/test/infra/async_utils.py
@@ -201,3 +201,12 @@ class EventLoopThread(ABC):
 
             # Clear the flag, so it can be re-used by subclasses.
             self._ready = False
+
+    def join(self) -> None:
+        """
+        Wait for the event loop thread to exit.
+        """
+        self._thread.join()
+
+        if self._exception is not None:
+            raise self._exception
--- a/test/infra/cchost.py
+++ b/test/infra/cchost.py
@@ -5,12 +5,13 @@ import argparse
 import asyncio
 import json
 import os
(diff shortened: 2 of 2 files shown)


letmaik:
When we were still using CCF's sandbox script in `start.sh` it was possible to add a `-d <node_idx>` flag which then printed out the gdb command that has to be manually run (in another terminal), waited for key press, and then continued with network setup etc. Apart from printing the gdb command it also wrote a file to `/tmp/vscode-gdb.sh` which allowed to start debugging directly from VS Code with launch configurations in the `.vscode/` folder.

See also:
- https://github.com/microsoft/CCF/pull/1444
- https://github.com/microsoft/CCF/pull/1650
- https://github.com/microsoft/scitt-ccf-ledger/blob/6d4a136cc814b6e995ea0a4a21f7e6c73e95bb55/.vscode/launch.json

I'm mentioning this just to give some context and ideally we'd want to restore the VS Code debugging option somehow. Do you think this PR is the right direction to do that eventually?

PR #112, opened 2023-02-20
Add integration tests for the DID resolver.

plietar:
Some of the resolver's functionality depends on the host's system time.
In order to test these parts, the test infrastucture now uses the
libfaketime utility to intercept library calls to time functions, and
allow us to introduce an artificial time shift.

A lot of the resolver's error handling cannot yet be tested, because
errors aren't surfaced through the API. When fixing https://github.com/microsoft/scitt-ccf-ledger/issues/108, we'll be able
to add a lot more tests.


letmaik (approved):
This is great!

PR #125, opened 2023-02-23
Jp1/attested errors

darracott:
Addresses https://github.com/microsoft/scitt-ccf-ledger/issues/105

### DIFF
--- a/app/fetch-did-web-doc.py
+++ b/app/fetch-did-web-doc.py
@@ -232,7 +232,6 @@ def process_requests(url, unattested):
                 retry = refetchable(result)
                 tries += 1
 
-
             # Send back the result to the callback URL.
             callback_data = {"result": json.loads(result)}
             body = json.dumps(callback_data).encode("utf-8")
--- a/app/fetch-did-web-doc.py
+++ b/app/fetch-did-web-doc.py
@@ -14,6 +14,7 @@ from urllib.request import Request, urlopen
 
 CONNECT_TIMEOUT = 5
 HTTP_RETRIES = 5
+FETCH_RETRIES = 5
 HTTP_DEFAULT_RETRY_AFTER = 1
 
 AFETCH_DIR = "/tmp/scitt"
@@ -26,8 +27,11 @@ def fetch_unattested(url, nonce):
         {
             "url": url,
             "nonce": nonce,
-            "status_code": response["status_code"],
-            "body": base64.b64encode(response["body"]).decode("utf-8"),
+            "result": {
+                "status": response["status_code"],
+                "body": base64.b64encode(response["body"]).decode("utf-8"),
+                "cert_chain": [],
+            },
         }
     ).encode("utf-8")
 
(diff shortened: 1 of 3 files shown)


test/test_resolution.py
```diff
@@ -322,6 +322,26 @@ def test_dont_cache_error(
         receipt = client.submit_claim(claim).receipt
         verify_receipt(claim, trust_store, receipt)
 
+    def test_fetch_error(
+        self,
+        client: Client,
+        did_web: DIDWebServer,
+    ):
+        """
+        Test that the ledger does not cache failed resolutions.
```
darracott, line 331: Need to update this description

app/fetch-did-web-doc.py
```diff
@@ -26,8 +27,11 @@ def fetch_unattested(url, nonce):
         {
             "url": url,
             "nonce": nonce,
-            "status_code": response["status_code"],
-            "body": base64.b64encode(response["body"]).decode("utf-8"),
+            "result": {
+                "status": response["status_code"],
+                "body": base64.b64encode(response["body"]).decode("utf-8"),
+                "cert_chain": [],
+            },
```
plietar, line 34: This is going to break the unattested parser in the ledger, and also doesn't work with refetchable. Given that `request` already does retries on http errors, there's no need to try and unify the two code paths together at this point. We can do that when we have the "virtual" attested-fetch, and then we'll be able to simplify this file.

app/fetch-did-web-doc.py
```diff
@@ -120,10 +143,13 @@ def request(url, data=None, headers=None):
...
-        result = fetch_unattested(url, nonce)
-    else:
-        result = fetch_attested(url, nonce)
+    fetch = fetch_unattested if unattested else fetch_attested
+    retry = True
+    tries = 0
+    while retry and tries < FETCH_RETRIES:
+        result = fetch(url, nonce)
+        retry = refetchable(result)
+        tries += 1
```
plietar, line 152: `fetch_unattested` uses `request`, which already has its own retry loop. IMO the retry loop here should be pushed into the fetch_attested function.

app/fetch-did-web-doc.py
darracott: Reverted the unattested changes

app/fetch-did-web-doc.py
darracott: I've updated this to only retry the attested fetch - I do think it's more clear to keep this level of retry here given we are inspecting the output of attested fetch to decide whether to repeat it, i.e. attested-fetch is completing and we are then repeating it. 
The attested-fetch function already has retry logic to deal with network issues, so potentially where attested-fetch is attesting to a curl error we could skip the higher level retries.

test/test_resolution.py
```diff
@@ -318,6 +318,23 @@ def test_dont_cache_error(
         receipt = client.submit_claim(claim).receipt
         verify_receipt(claim, trust_store, receipt)
 
+    def test_fetch_error(
```
plietar, line 321: You can add
```python
@pytest.mark.skipif(os.getenv("PLATFORM") == "virtual")
```
to disable this when using unattested fetch.

app/fetch-did-web-doc.py
letmaik: > The attested-fetch function already has retry logic to deal with network issues, so potentially where attested-fetch is attesting to a curl error we could skip the higher level retries.

I think that's not accurate as network issues also lead to an attested fetch result (where `error` is present). Only when the enclave crashes for some reason would the existing retry logic in `fetch_attested` kick in now. I agree with @plietar that the retry loop should be inside `fetch_attested` but we can also do that in a follow-up PR.

letmaik (approved):

Issue #126, opened 2023-02-23
Log requests with failed auth


letmaik:
Auth errors fully skip any endpoint handlers including our tracing adapter. It would still be good to log auth errors somehow, maybe by extending https://github.com/microsoft/scitt-ccf-ledger/blob/main/app/src/configurable_auth.h.

letmaik:
Done in #130.

PR #130, opened 2023-02-28
add logs for auth-errors

ryazhang-microsoft:
https://github.com/microsoft/scitt-ccf-ledger/issues/126

1. add logs for failed auth:  ClientRequestId=xx ErrorMessage=xxx
2. enhance the submit_claim function at pyscitt/client.py - allow custom headers

### DIFF
--- a/app/src/configurable_auth.h
+++ b/app/src/configurable_auth.h
@@ -88,7 +88,7 @@ namespace scitt
     {
       // CCF returns any errors in the auth policy with a 401 status
       CCF_APP_INFO(
-        "ClientRequestId={} Verb={} URL={} Status_code=401",
+        "ClientRequestId={} Verb={} URL={} Status=401",
         ctx->get_request_header("x-ms-client-request-id").value_or(""),
         ctx->get_request_verb().c_str(),
         ctx->get_request_url());
--- a/app/src/configurable_auth.h
+++ b/app/src/configurable_auth.h
@@ -27,8 +27,11 @@ namespace scitt
     {
       auto identity = JwtAuthnPolicy::authenticate(tx, ctx, error_reason);
       if (!identity)
-      { 
-        SCITT_INFO("ClientRequestId={} ErrorMessage={}", ctx->get_request_header(CLIENT_REQUEST_ID_HEADER).value_or(""), error_reason);
+      {
+        SCITT_INFO(
+          "ClientRequestId={} ErrorMessage={}",
+          ctx->get_request_header(CLIENT_REQUEST_ID_HEADER).value_or(""),
+          error_reason);
         return nullptr;
       }
 
@@ -45,7 +48,10 @@ namespace scitt
       if (!required_claims.is_object())
       {
         error_reason = "JWT authentication is not enabled";
-        SCITT_INFO("ClientRequestId={} ErrorMessage={}", ctx->get_request_header(CLIENT_REQUEST_ID_HEADER).value_or(""), error_reason);
+        SCITT_INFO(
+          "ClientRequestId={} ErrorMessage={}",
+          ctx->get_request_header(CLIENT_REQUEST_ID_HEADER).value_or(""),
+          error_reason);
         return nullptr;
       }
 
@@ -55,7 +61,10 @@ namespace scitt
       }
       else
       {
-        SCITT_INFO("ClientRequestId={} ErrorMessage={}", ctx->get_request_header(CLIENT_REQUEST_ID_HEADER).value_or(""), error_reason);
+        SCITT_INFO(
+          "ClientRequestId={} ErrorMessage={}",
+          ctx->get_request_header(CLIENT_REQUEST_ID_HEADER).value_or(""),
+          error_reason);
         return nullptr;
       }
     }
--- a/app/src/configurable_auth.h
+++ b/app/src/configurable_auth.h
@@ -4,6 +4,7 @@
 #pragma once
 
 #include "kv_types.h"
+#include "tracing.h"
 
 #include <ccf/common_auth_policies.h>
 
@@ -26,7 +27,8 @@ namespace scitt
     {
       auto identity = JwtAuthnPolicy::authenticate(tx, ctx, error_reason);
       if (!identity)
-      {
+      { 
+        SCITT_INFO("ClientRequestId={} ErrorMessage={}", ctx->get_request_header(CLIENT_REQUEST_ID_HEADER).value_or(""), error_reason);
         return nullptr;
       }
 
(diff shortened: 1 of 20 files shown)


pyscitt/pyscitt/client.py
```diff
@@ -369,14 +369,13 @@ def submit_claim(
...
     @overload
     def submit_claim(
-        self, claim: bytes, *, skip_confirmation: Literal[True]
+        self, claim: bytes, *, skip_confirmation: Literal[True], headers = {"Content-Type": "application/cose"}
     ) -> PendingSubmission:
         ...
 
     def submit_claim(
-        self, claim: bytes, *, skip_confirmation=False
+        self, claim: bytes, *, skip_confirmation=False, headers = {"Content-Type": "application/cose"}
```
plietar, line 377: Rather than having this as a default value, I think we should take whatever the caller passed and add the "content-type" header. Otherwise, a call with `submit_claim(..., headers={  "x-ms-client-request-id": "foo" })` would be missing the content type.

ryazhang-microsoft: Good point. Change to update instead of replace header.

app/src/configurable_auth.h
```diff
@@ -27,6 +28,10 @@ namespace scitt
       auto identity = JwtAuthnPolicy::authenticate(tx, ctx, error_reason);
       if (!identity)
       {
+        SCITT_INFO(
```
letmaik, line 31: This code path would also be reached when the configuration allows unauthenticated access (then the `ConfigurableEmptyAuthnPolicy` would succeed), so the check for that (see below) has to be moved up, otherwise we log errors when there haven't been any.

app/src/configurable_auth.h
```diff
@@ -27,6 +28,10 @@ namespace scitt
...
+          "ClientRequestId={} ErrorMessage={}",
```
letmaik, line 32: I think the format should match the log messages emitted for endpoint-originating errors, including the URL etc.

app/src/configurable_auth.h
```diff
@@ -4,6 +4,7 @@
 #pragma once
 
 #include "kv_types.h"
+#include "tracing.h"
```
letmaik, line 7: Given that logging in authn policy classes runs before the actual endpoint starts there's not really a reason to use our custom logging macros here (as it wouldn't add request id or client request id anyway). By using `CCF_..` we would avoid this extra include and also the `inline`s everywhere.

app/src/configurable_auth.h
ryazhang-microsoft: It seem the ConfigurableEmptyAuthnPolicy is cehck before ConfigurableJwtAuthnPolicy.
see main.cpp 282
const ccf::AuthnPolicies authn_policy = {
        std::make_shared<ConfigurableEmptyAuthnPolicy>(),
        std::make_shared<ConfigurableJwtAuthnPolicy>(),
      };
According to the docs in ccf, any single successful check is sufficient to grant access. And it seem the check is in order of the authn_policy vector.

app/src/configurable_auth.h
ryazhang-microsoft: add verb and url

app/src/configurable_auth.h
ryazhang-microsoft: fixed

app/src/configurable_auth.h
letmaik: You're right, the order matters of course.

pyscitt/pyscitt/client.py
```diff
@@ -363,20 +363,24 @@ def get_did_document(self, did: str) -> dict:
...
 
     def submit_claim(
-        self, claim: bytes, *, skip_confirmation=False
+        self, claim: bytes, *, skip_confirmation=False, headers=None
     ) -> Union[Submission, PendingSubmission]:
-        headers = {"Content-Type": "application/cose"}
+        request_header = {"Content-Type": "application/cose"}
+
+        if headers:
+            request_header.update(headers)
```
letmaik, line 382: You don't use `request_header` below. It might be cleaner to stick with a single variable. Maybe something like:
```py
if headers is None:
  headers = {}
headers["content-type"] = "..."
```

app/src/configurable_auth.h
```diff
@@ -27,6 +28,12 @@ namespace scitt
       auto identity = JwtAuthnPolicy::authenticate(tx, ctx, error_reason);
       if (!identity)
       {
+        CCF_APP_INFO(
+          "ClientRequestId={} Verb={} URL={} ErrorMessage={}",
```
letmaik, line 32: We should add `Status=401` here with a comment saying that CCF returns any errors in the auth policy with a 401 status.

letmaik, line 32: The error message is unstructured and will contain spaces, so it seems a bit odd to pretend it's parseable as key-value pair somehow. In the `generic_error_adapter` we log `... Code={code} {msg}` for internal errors and `... Code={code}` for user errors. Do you think we should stick to the same and emit a second log message like below?

`ClientRequestId={id} Code=InvalidAuthenticationInfo {msg}`

CCF uses `InvalidAuthenticationInfo` as code for 401 errors.

@plietar Any other ideas?

pyscitt/pyscitt/client.py
plietar: @letmaik this has the side effect of modifying the caller's dictionary, which could end up being suprising.

app/src/configurable_auth.h
plietar: >@plietar Any other ideas?

Only bad ones: Structure log messages as JSON.

Otherwise what you suggested with the parseable Code and trailing message is fine I think.

app/src/configurable_auth.h
```diff
@@ -78,6 +82,21 @@ namespace scitt
       }
       return true;
     }
+
+    static void log_auth_error(
+      const std::shared_ptr<ccf::RpcContext>& ctx, std::string& error_reason)
+    {
+      // CCF returns any errors in the auth policy with a 401 status
+      CCF_APP_INFO(
+        "ClientRequestId={} Verb={} URL={} Status_code=401",
```
letmaik, line 91:
```suggestion
        "ClientRequestId={} Verb={} URL={} Status=401",
```

We should use a consistent format here, see https://github.com/microsoft/scitt-ccf-ledger/blob/3ceb7d750f27e5ee8ce95207b30f8253919b6f51/app/src/tracing.h#L108.

pyscitt/pyscitt/client.py
letmaik: @ryazhang-microsoft You marked this as resolved but the `request_header` variable you introduced is still not used anywhere. Please also confirm where this new keyword argument is used.

PR #131, opened 2023-03-02
Use a virtual build of attested-fetch on non-SGX.

plietar:
Attested fetch can now be built in a "virtual" mode, where it runs directly on the host instead of in an SGX enclave. We can use this in our own virtual builds, instead of the previous "unattested fetch", which was implemented in Python.

Of course, despite being called *attested* fetch, the virtual mode lacks any kind of attestation and is not any more or less secure than the previous unattested fetch.

This makes the two environments more uniform, and removes most of the code difference in the ledger. The only difference left is that SGX builds of the ledger will check the hardware attestation, whereas the virtual mode doesn't.

As an added benefit, features that were missing in "unattested fetch", namely checking of TLS certificates and reporting HTTP errors, are now available even in virtual builds.


letmaik (approved):

letmaik:
CI is failing:
> /.pipelines/common.yml (Line: 54, Col: 11): Unexpected value 'displayName'
551

letmaik:
Yes, I manually went into Azure Pipelines and searched for a failed run for this branch. Unfortunately, it doesn't report YAML errors back to GitHub and just leaves it pending.

PR #132, opened 2023-03-03
Remove verify work-arounds for Notary style claims

darracott:
Supersedes PR: https://github.com/microsoft/scitt-ccf-ledger/pull/107 
Addresses issue: https://github.com/microsoft/scitt-ccf-ledger/issues/101

 Replaces workaround signature verification code with t_cose for Notary style claims.

To discuss: Adds checks (for the notary validation case) to ensure critical params are known and present in the protected header.

### DIFF
--- a/app/src/cose.h
+++ b/app/src/cose.h
@@ -202,12 +202,12 @@ namespace scitt::cose
     // parameter in the context of a profile.
     bool is_known(
       const std::variant<int64_t, std::string>& label,
-      std::set<std::variant<int64_t, std::string>> profile_paramters) const
+      std::set<std::variant<int64_t, std::string>> profile_parameters) const
     {
       if (
         BASIC_HEADER_PARAMS.contains(label) ||
         EXTRA_HEADER_PARAMS.contains(label) ||
-        profile_paramters.contains(label))
+        profile_parameters.contains(label))
       {
         return true;
       }
--- a/app/src/cose.h
+++ b/app/src/cose.h
@@ -645,17 +645,12 @@ namespace scitt::cose
     t_cose_sign1_verify_ctx verify_ctx;
 
     // Do some preliminary decoding, to get the header parameters and potential
-    // auxiliary buffer size. T_COSE_OPT_UNKNOWN_CRIT_ALLOWED
+    // auxiliary buffer size.
     t_cose_parameters params;
-    uint32_t prelim_options;
+    uint32_t prelim_options = T_COSE_OPT_TAG_REQUIRED | T_COSE_OPT_DECODE_ONLY;
     if (allow_unknown_crit)
     {
-      prelim_options = T_COSE_OPT_TAG_REQUIRED | T_COSE_OPT_DECODE_ONLY |
-        T_COSE_OPT_UNKNOWN_CRIT_ALLOWED;
-    }
-    else
-    {
-      prelim_options = T_COSE_OPT_TAG_REQUIRED | T_COSE_OPT_DECODE_ONLY;
+      prelim_options |= T_COSE_OPT_UNKNOWN_CRIT_ALLOWED;
     }
     t_cose_sign1_verify_init(&verify_ctx, prelim_options);
     t_cose_err_t error =
@@ -680,14 +675,10 @@ namespace scitt::cose
     EVP_PKEY* evp_key = key.get_evp_pkey();
     cose_key.k.key_ptr = evp_key;
 
-    uint32_t options;
+    uint32_t options = T_COSE_OPT_TAG_REQUIRED;
     if (allow_unknown_crit)
     {
-      options = T_COSE_OPT_TAG_REQUIRED | T_COSE_OPT_UNKNOWN_CRIT_ALLOWED;
-    }
-    else
-    {
-      options = T_COSE_OPT_TAG_REQUIRED;
+      options |= T_COSE_OPT_UNKNOWN_CRIT_ALLOWED;
     }
     t_cose_sign1_verify_init(&verify_ctx, options);
     t_cose_sign1_set_verification_key(&verify_ctx, cose_key);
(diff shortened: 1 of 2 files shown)


app/src/cose.h
```diff
@@ -76,6 +96,91 @@ namespace scitt::cose
     std::optional<int64_t> notary_authentic_signing_time;
     std::optional<int64_t> notary_expiry;
 
+    bool is_present(const std::variant<int64_t, std::string>& label) const
```
darracott, line 99: Surely there is a better way to do this but I couldn't find a simple way to access a struct's fields by name

app/src/cose.h
```diff
@@ -91,6 +196,36 @@ namespace scitt::cose
...
       return false;
     }
+
+    // Returns a bool representing whether the input label is a known
+    // parameter in the context of a notary profile.
+    bool is_known_in_notary(
+      const std::variant<int64_t, std::string>& label) const
+    {
+      // List of known header parameters in a notary profile
+      auto notary_known_params(BASIC_HEADER_PARAMS);
```
darracott, line 206: Is there a nicer way to combine lists?

app/src/cose.h
```diff
@@ -700,10 +645,19 @@ namespace scitt::cose
...
     // Do some preliminary decoding, to get the header parameters and potential
-    // auxiliary buffer size.
+    // auxiliary buffer size. T_COSE_OPT_UNKNOWN_CRIT_ALLOWED
     t_cose_parameters params;
-    t_cose_sign1_verify_init(
-      &verify_ctx, T_COSE_OPT_TAG_REQUIRED | T_COSE_OPT_DECODE_ONLY);
+    uint32_t prelim_options;
+    if (allow_unknown_crit)
+    {
+      prelim_options = T_COSE_OPT_TAG_REQUIRED | T_COSE_OPT_DECODE_ONLY |
```
letmaik, line 653: I would avoid the duplication and in this line just do `prelim_options |= T_COSE_OPT_UNKNOWN_CRIT_ALLOWED;` The assignment in the else branch would be the initial assignment.

app/src/cose.h
plietar: Not really,  but in this particular case you don't actually need the combined list, instead you could search in all three lists.

If you make them `std::std`s instead of vectors, this is even shorter as `BASIC_HEADER_PARAMS.contains(label) || EXTRA_HEADER_PARAMS.contains(label) || NOTARY_HEADER_PARAMS.contains(label)`

app/src/verifier.h
```diff
@@ -186,10 +186,26 @@ namespace scitt::verifier
...
+          throw cose::COSEDecodeError("Unknown parameter found in crit");
+        }
+        else
+        {
+          if (!phdr.is_present(crit_param))
+          {
+            SCITT_INFO(
+              "Critical param {} missing from protected header", crit_param);
+            throw cose::COSEDecodeError(
+              "Critial param missing from protected header");
```
letmaik, line 205:
```suggestion
              "Critical parameter missing from protected header");
```

app/src/verifier.h
```diff
@@ -326,9 +342,7 @@ namespace scitt::verifier
           // Verify signature.
           try
           {
-            // TODO: replace with cose::verify() once the t_cose workarounds are
-            // removed.
-            cose::notary_verify(data, phdr, key);
+            cose::verify(data, key, true);
```
letmaik, line 345:
```suggestion
            cose::verify(data, key, /* allow_unknown_crit */ true);
```

Also, add a short comment above saying where critical parameters are checked.

app/src/verifier.h
```diff
@@ -186,10 +186,26 @@ namespace scitt::verifier
...
+        {
+          SCITT_INFO("Unknown critical parameter: {}", crit_param);
+          throw cose::COSEDecodeError("Unknown parameter found in crit");
+        }
+        else
+        {
+          if (!phdr.is_present(crit_param))
+          {
+            SCITT_INFO(
+              "Critical param {} missing from protected header", crit_param);
```
letmaik, line 203:
```suggestion
              "Critical parameter {} missing from protected header", crit_param);
```

app/src/cose.h
```diff
@@ -700,10 +645,19 @@ namespace scitt::cose
     t_cose_sign1_verify_ctx verify_ctx;
 
     // Do some preliminary decoding, to get the header parameters and potential
-    // auxiliary buffer size.
+    // auxiliary buffer size. T_COSE_OPT_UNKNOWN_CRIT_ALLOWED
```
letmaik, line 648:
```suggestion
    // auxiliary buffer size.
```

app/src/cose.h
```diff
@@ -91,6 +196,36 @@ namespace scitt::cose
       }
       return false;
     }
+
+    // Returns a bool representing whether the input label is a known
+    // parameter in the context of a notary profile.
+    bool is_known_in_notary(
```
letmaik, line 202: I don't think profile-specific code (apart from decoding) should be in `cose.h`. Could this move into `verifier.h`?

app/src/cose.h
```diff
@@ -91,6 +197,22 @@ namespace scitt::cose
       }
       return false;
     }
+
+    // Returns a bool representing whether the input label is a known
+    // parameter in the context of a notary profile.
```
letmaik, line 202:
```suggestion
    // parameter in the context of a profile.
```

app/src/cose.h
```diff
@@ -91,6 +197,22 @@ namespace scitt::cose
...
+    bool is_known(
+      const std::variant<int64_t, std::string>& label,
+      std::set<std::variant<int64_t, std::string>> profile_paramters) const
+    {
+      if (
+        BASIC_HEADER_PARAMS.contains(label) ||
+        EXTRA_HEADER_PARAMS.contains(label) ||
+        profile_paramters.contains(label))
```
letmaik, line 210:
```suggestion
      std::set<std::variant<int64_t, std::string>> profile_parameters) const
    {
      if (
        BASIC_HEADER_PARAMS.contains(label) ||
        EXTRA_HEADER_PARAMS.contains(label) ||
        profile_parameters.contains(label))
```

letmaik (approved):

PR #138, opened 2023-03-09
Fix load test

### LINKED ISSUE #37: Revive load testing and run in daily CI

letmaik:
Closes #37.

### DIFF
--- a/test/load_test/locustfile.py
+++ b/test/load_test/locustfile.py
@@ -3,138 +3,50 @@
 # type: ignore
 
 import random
-import subprocess
-import sys
-import tempfile
-import traceback
 from pathlib import Path
 
-from locust import FastHttpUser, events, task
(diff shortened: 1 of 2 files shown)


test/load_test/locustfile.py
```diff
@@ -3,138 +3,50 @@
 # type: ignore
```
plietar, line 3: Can this be removed now?

plietar (approved):

test/load_test/locustfile.py
letmaik: Yep, done.

PR #139, opened 2023-03-13
Enhance devcontainer config

letmaik:
- Install missing libcurl4-openssl-dev package
- Set minimum CPU count to let GitHub pick a reasonable default
- Set `PLATFORM=virtual` to avoid having to type it when invoking build/test scripts

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,12 +1,18 @@
 {
-  "name": "Development Container for CCF C++ Apps",
+  "name": "Development Container for scitt-ccf-ledger",
   "dockerFile": "Dockerfile",
-  "extensions": [
-    "ms-vscode.cpptools"
-  ],
+
+  "customizations": {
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -1,12 +1,23 @@
...
+      ]
+    }
+  },
   "features": {
     "ghcr.io/devcontainers/features/docker-from-docker:1": {
       "version": "20.10.8"
     }
-  }
+  },
+  "onCreateCommand": "apt-get update && apt-get install -y libcurl4-openssl-dev"
```
plietar, line 22: Should this live in the `Dockerfile` instead? I'm not sure if it makes much difference tbh.

letmaik: The outcome is the same, but yes, it makes more sense to put it in the `Dockerfile`, forgot about that.

plietar (approved):

PR #144, opened 2023-07-21
Wait for cchost config to be available

andpiccione:
This PR modifies the SCITT Dockerfiles to include a simple script that allows waiting for the `cchost` config file to be available and exit if the condition is not met within a certain timeout. 

This is to support the current deployment of SCITT in Kubernetes using the CCF orchestration layer without sidecars. This configuration will eventually be removed once we upgrade SCITT to CCF 4.0.3, which supports a new `config-timeout` option to achieve a similar goal.

### DIFF
--- a/docker/run-dev.sh
+++ b/docker/run-dev.sh
@@ -90,7 +90,7 @@ docker run --name "$CONTAINER_NAME" \
     -d \
     "${DOCKER_FLAGS[@]}" \
     -v "$VOLUME_NAME":/host \
-    "$DOCKER_TAG" "--config /host/dev-config.json"
+    "$DOCKER_TAG" --config /host/dev-config.json
 
 echo "Setting up python virtual environment."
 if [ ! -f "venv/bin/activate" ]; then
--- a/docker/run-dev.sh
+++ b/docker/run-dev.sh
@@ -90,7 +90,7 @@ docker run --name "$CONTAINER_NAME" \
     -d \
     "${DOCKER_FLAGS[@]}" \
     -v "$VOLUME_NAME":/host \
-    "$DOCKER_TAG" --config /host/dev-config.json
+    "$DOCKER_TAG" "--config /host/dev-config.json"
 
 echo "Setting up python virtual environment."
 if [ ! -f "venv/bin/activate" ]; then
--- a/docker/start-app.sh
+++ b/docker/start-app.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+echo "Starting app"
+config_arg=$1
+
+cchost --check ${config_arg}
+code=$?
+while [[ $code -ne 0 ]]; do
+    echo "Waiting for configuration file to be ready..."
+    sleep 10
+    cchost --check ${config_arg}
+    code=$?
+done
+
+echo "Running cchost from $(pwd)"
+stdbuf -o L cchost ${config_arg}
--- a/docker/start-app.sh
+++ b/docker/start-app.sh
@@ -5,12 +5,12 @@
 # the SCITT container is deployed in Kubernetes without any orchestration sidecars.
 
 WAIT_TIME_SEC=3
-TIMEOUT_SEC=300
+TIMEOUT_SEC=5
 
 echo "Starting app"
-config_arg=$1
+config_arg=${1:?first argument is the configuration argument to pass to cchost (e.g., "--config /path/to/config.json"))}
 
-cchost --check ${config_arg}
+cchost --check "${config_arg}"
 code=$?
 
 start_time=$(date +%s)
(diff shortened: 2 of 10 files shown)


docker/start-app.sh
```diff
@@ -0,0 +1,33 @@
...
+# a cchost "config-timeout" option to accomplish a similar result. This logic is required when
+# the SCITT container is deployed in Kubernetes without any orchestration sidecars.
+
+WAIT_TIME_SEC=3
+TIMEOUT_SEC=300
+
+echo "Starting app"
+config_arg=$1
+
+cchost --check ${config_arg}
```
ivarprudnikov, line 13: quote `${config_arg}` like `"${config_arg}"`

docker/start-app.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# This configuration is a temporary workaround until we migrate to CCF 4.0.3, which supports
+# a cchost "config-timeout" option to accomplish a similar result. This logic is required when
+# the SCITT container is deployed in Kubernetes without any orchestration sidecars.
+
+WAIT_TIME_SEC=3
+TIMEOUT_SEC=300
+
+echo "Starting app"
+config_arg=$1
```
ivarprudnikov, line 11: I do not understand what this argument is, should you force it like:

```
config_arg=${1:?first argument must be the configuration path to pass to the cchost}
```

docker/start-app.sh
```diff
@@ -0,0 +1,33 @@
...
+
+cchost --check ${config_arg}
+code=$?
+
+start_time=$(date +%s)
+
+while [[ $code -ne 0 ]]; do
+    
+    # Exit if timeout is reached
+    if [[ $(($(date +%s) - $start_time)) -ge $TIMEOUT_SEC ]]; then
```
ivarprudnikov, line 21: simplify and move out `$(date +%s)` into a local variable, there is also an issue with the use of `$` in `$(())` documented in https://www.shellcheck.net/wiki/SC2004

andpiccione:
Ah, I had the shellcheck VS Code extension in my local VS Code, but forgot to add it to the GH codespace that I'm using to work on this repo :D Thanks for the reminder!

docker/start-app.sh
andpiccione: Addressed

docker/start-app.sh
andpiccione: Addressed

docker/start-app.sh
andpiccione: Done, forgot to install the shellcheck extension in the GH codespace to catch these type of issues! 😄

docker/start-app.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# This configuration is a temporary workaround until we migrate to CCF 4.0.3, which supports
+# a cchost "config-timeout" option to accomplish a similar result. This logic is required when
+# the SCITT container is deployed in Kubernetes without any orchestration sidecars.
+
+WAIT_TIME_SEC=3
+TIMEOUT_SEC=5
+
+echo "Starting app"
+config_arg=${1:?first argument is the configuration argument to pass to cchost (e.g., "--config /path/to/config.json"))}
```
ivarprudnikov, line 11: Actually by looking into how the parameters passed it seems that his will not work, e.g.
https://github.com/microsoft/scitt-ccf-ledger/blob/main/docker/run-dev.sh#L93

it passes them as separate values to `docker run .... --config /path/to/config.json` so here $1 will be `--config` and $2 will be `/path/to/config.json`

andpiccione: Good point. For some reason it was working fine when I tested in from the cluster(because I think it is passed a single string), but indeed it doesn't work when tested locally. I just pushed a fix to get all args.

docker/run-dev.sh
```diff
@@ -90,7 +90,7 @@ docker run --name "$CONTAINER_NAME" \
     -d \
     "${DOCKER_FLAGS[@]}" \
     -v "$VOLUME_NAME":/host \
-    "$DOCKER_TAG" --config /host/dev-config.json
+    "$DOCKER_TAG" "--config /host/dev-config.json"
```
ivarprudnikov, line 93: no need to quote here

docker/start-app.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# This configuration is a temporary workaround until we migrate to CCF 4.0.3, which supports
+# a cchost "config-timeout" option to accomplish a similar result. This logic is required when
+# the SCITT container is deployed in Kubernetes without any orchestration sidecars.
+
+WAIT_TIME_SEC=3
+TIMEOUT_SEC=300
+
+echo "Starting app"
+args=$(printf '%s ' "$@") # concatenate all input cchost arguments into a single string
```
ivarprudnikov, line 11: should not be necessary, using `"$@"` subsequently would work

docker/run-dev.sh
andpiccione: Ops, leftover for test. Thanks!

docker/start-app.sh
andpiccione: Done

ivarprudnikov (approved):

PR #146, opened 2023-08-28
Adds new GitHub action to run unit tests

ivarprudnikov:
The main idea is to reduce the dependence on Azure resources and make sure tests are always run. A good example is a current Azure resource config which is difficult to setup due to the internal requirements and the nature of SGX. The resources were recently removed which stopped the tests from running even in virtual mode. Even though they are being added back in, we need to make sure the dependence is derisked.

New GitHub action:
- Run a build in virtual mode
- Run unit tests
- Run functional tests
- Check code formatting

### DIFF
--- a/.github/workflows/build-test-virtual.yml
+++ b/.github/workflows/build-test-virtual.yml
@@ -1,0 +1,16 @@
+name: "Build/test (virtual)"
+on: [push]
+jobs:
+  build-test:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    container:
+      image: mcr.microsoft.com/ccf/app/dev:3.0.9-virtual
+      env:
+        PLATFORM: virtual
+    steps:
+      - uses: actions/checkout@v3
+      - run: ./build.sh
+      - run: ./run_unit_tests.sh
--- a/.github/workflows/build-test-virtual.yml
+++ b/.github/workflows/build-test-virtual.yml
@@ -12,5 +12,8 @@ jobs:
         PLATFORM: virtual
     steps:
       - uses: actions/checkout@v3
+      # Work-around for https://github.com/actions/runner/issues/2033
+      - name: Work around git safe.directory in container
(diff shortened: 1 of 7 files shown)


.github/workflows/build-test-virtual.yml
```diff
@@ -0,0 +1,34 @@
...
+    container:
+      image: mcr.microsoft.com/ccf/app/dev:3.0.9-virtual
+      env:
+        # Helps to distinguish between CI and local builds.
+        SCITT_CI: 1
+        # Additional CMake flags.
+        PLATFORM: virtual
+        CMAKE_BUILD_TYPE: Debug
+        ENABLE_PREFIX_TREE: ON
+        ENABLE_CLANG_TIDY: ON
```
andpiccione, line 18: Do you think it could make sense to enable performance tests as well through the corresponding environment variable?

.github/workflows/build-test-virtual.yml
```diff
@@ -0,0 +1,34 @@
+name: "Build/test (virtual)"
+on: [push]
+jobs:
+  build-test:
+    runs-on: ubuntu-20.04
```
andpiccione, line 5: nit: could this run on ubuntu 22.04 (just to use the latest LTS release) ?

andpiccione (approved):
LGTM!

.github/workflows/build-test-virtual.yml
ivarprudnikov: no we cannot use v22 until CCF fully tests and supports it

.github/workflows/build-test-virtual.yml
ivarprudnikov: They are/were enabled in Azure pipelines but only on the proper SGX hardware https://github.com/microsoft/scitt-ccf-ledger/blob/main/.pipelines/azure-pipelines.yml I was thinking about enabling it here but is feels like overkill and a bit misleading to do it in the PRs using virtual

Issue #141, opened 2023-03-22
Update to CCF 3.0.10


letmaik:
https://github.com/microsoft/CCF/releases/tag/ccf-3.0.10

PR #147, opened 2023-08-29
Upgrade to CCF 3.0.13

vimauro:
This PR upgrades CCF to version 3.0.13. This is an intermediate step before moving to 4.x/5.x .

Upgraded to `clang-11` for SGX and `clang-15` for Virtual builds since those are CCF 3.0.13 requirements (https://github.com/microsoft/CCF/releases/tag/ccf-3.0.13)

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -6,8 +6,13 @@ cmake_minimum_required(VERSION 3.16)
 if((NOT CMAKE_CXX_COMPILER)
    AND "$ENV{CXX}" STREQUAL ""
 )
-  set(CMAKE_CXX_COMPILER clang++-11)
-  message("Set CMAKE_CXX_COMPILER to clang++-11")
+  if (${COMPILE_TARGET} STREQUAL "sgx")
+    set(CMAKE_CXX_COMPILER clang++-11)
+    message("Set CMAKE_CXX_COMPILER to clang++-11")
+  else()
+    set(CMAKE_CXX_COMPILER clang++-15)
+    message("Set CMAKE_CXX_COMPILER to clang++-15")  
+  endif()
 endif()
 
 project(scitt


build.sh
```diff
@@ -9,8 +9,8 @@ PLATFORM=${PLATFORM:-sgx}
 CCF_UNSAFE=${CCF_UNSAFE:-OFF}
 ENABLE_PREFIX_TREE=${ENABLE_PREFIX_TREE:-OFF}
 BUILD_TESTS=${BUILD_TESTS:-ON}
-CC=${CC:-clang-10}
-CXX=${CXX:-clang++-10}
+CC=${CC:-clang-11}
+CXX=${CXX:-clang++-11}
```
andpiccione, line 13: Would these need to be set to either clang-11 or clang-15 depending on the `PLATFORM` variable? Since clang-11 is supported for SGX and clang-15 is supported for virtual

andpiccione (commented):
Overall, lgtm. But I believe we cannot safely merge this PR until we re-enable the build for SGX to validate that all test scenarios pass in every platform

scripts/check-format.sh
```diff
@@ -32,9 +32,9 @@ file_name_regex="^[[:lower:]0-9_]+$"
 unformatted_files=""
 badly_named_files=""
 for file in $(git ls-files "$@" | grep -e '\.h$' -e '\.hpp$' -e '\.cpp$' -e '\.c$' -e '\.proto$'); do
-  if ! clang-format-10 -n -Werror -style=file "$file"; then
+  if ! clang-format-11 -n -Werror -style=file "$file"; then
```
andpiccione, line 35: Since you have upgraded clang-format too, I think you would need to format all the files with clang-format-11, since some formatting rules might have changed from clang-format-10.

vimauro:
Addressed #141 by upgrading to 3.0.13

build.sh
vimauro: Good point, added a conditional check based on PLATFORM

.github/workflows/build-test-virtual.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       # Work-around for https://github.com/actions/runner/issues/2033
       - name: Work around git safe.directory in container
         run: chown -R $(id -u):$(id -g) $PWD
-      - run: apt-get update && apt-get install -y libcurl4-openssl-dev faketime clang-tidy-10
+      - run: apt-get update && apt-get install -y libcurl4-openssl-dev faketime clang-tidy-12
```
ivarprudnikov, line 30: why are you changing this? Is it related to the move to 3.0.13?

build.sh
```diff
@@ -23,25 +21,30 @@ install_dir=/tmp/scitt
...
-ninja -C build/attested-fetch ${NINJA_FLAGS} --verbose
-ninja -C build/attested-fetch ${NINJA_FLAGS} install
 
 if [ "$PLATFORM" = "sgx" ]; then
     ATTESTED_FETCH_MRENCLAVE_HEX=$(/opt/openenclave/bin/oesign dump -e $install_dir/libafetch.enclave.so.signed | sed -n "s/mrenclave=//p")
+    CC=${CC:-clang-11}
+    CXX=${CXX:-clang++-11}
 elif [ "$PLATFORM" = "virtual" ]; then
     ATTESTED_FETCH_MRENCLAVE_HEX=""
+    CC=${CC:-clang-15}
```
ivarprudnikov, line 31: what is the reasoning for this change? why make it different from SGX?

ivarprudnikov (changes requested):

build.sh
vimauro: It's a requirement of CCF 3.0.13: [Releases · microsoft/CCF (github.com)](https://github.com/microsoft/CCF/releases)

 clang-11 for SGX and clang-15 for virtual

.github/workflows/build-test-virtual.yml
vimauro: not necessarily, just wanted to use a more recent version. The format errors currently flagged in the CI pipeline are occurring also with `clang-tidy-10` , I'm working on those rn

ivarprudnikov:
You'd need to change `clang-format-10` to `clang-format-11` in `check-format.sh` like in https://github.com/microsoft/CCF/blob/main/scripts/check-format.sh

vimauro:
> You'd need to change `clang-format-10` to `clang-format-11` in `check-format.sh` like in https://github.com/microsoft/CCF/blob/main/scripts/check-format.sh
> 
> Please also elaborate on the changes in the the issue description, e.g., why changing clang versions as well

Added more details in the description.
`clang-format-11` is giving me some `command not found ` errors, will keep it to `clang-format-10` for the moment if that's ok

PR #150, opened 2023-09-05
Updated Azure Pipelines pool

vimauro:
I have set up a new agent pool to be used by Azure Pipelines since the old one was removed. 
This PR makes use of the new pool and restores the _LOCUST_PEAK_USERS_ to 100 which is the original value while we investigate why 500 fails in SGX with no prefix tree.


ivarprudnikov:
Add a readme explaining how to setup the pipeline, there is a doc directory for this:
* how to create a pool
* how to link it in azure devops

ivarprudnikov (dismissed):
Add a readme explaining how to setup the pipeline, there is a doc directory for this:

how to create a pool
how to link it in azure devops

test/test_load.py
```diff
@@ -11,7 +11,7 @@
 from pyscitt.client import Client
 
 NUM_CLAIMS = 100
-LOCUST_PEAK_USERS = 500
+LOCUST_PEAK_USERS = 100
 LOCUST_USERS_SPAWN_RATE = 10
```
andpiccione, line 15: nit: since we have decreased the peak number of users, it might make sense to decrease a little the spawn rate (like 5).

andpiccione (approved):
lgtm

andpiccione (commented):
Should we now remove the [GitHub action to run virtual tests](https://github.com/microsoft/scitt-ccf-ledger/blob/main/.github/workflows/build-test-virtual.yml), that we recently added? The full build should now cover everything and it seems repetitive to keep both just for reducing dependency on Azure.

vimauro:
> Add a readme explaining how to setup the pipeline, there is a doc directory for this:
> 
> how to create a pool how to link it in azure devops

Those are msft internal instructions while Azure Pipelines is an App already installed at Organization level (https://github.com/apps/azure-pipelines)

vimauro:
> Should we now remove the [GitHub action to run virtual tests](https://github.com/microsoft/scitt-ccf-ledger/blob/main/.github/workflows/build-test-virtual.yml), that we recently added? The full build should now cover everything and it seems repetitive to keep both just for reducing dependency on Azure.

Makes sense to me. Thoughts @ivarprudnikov ?

test/test_load.py
vimauro: Thanks, decreased to 5

ivarprudnikov:
I would not remove Github action just yet as it is quite fast and acts as a
fallback in a case azure pipeline gets removed or similar.

PR #153, opened 2023-09-20
Steps to reproduce mrenclave

ivarprudnikov:
Following the addition of MRENCLAVE in the build logs this is a set of steps that will reproduce the same value.

### DIFF
--- a/docs/reproducibility.md
+++ b/docs/reproducibility.md
@@ -26,13 +26,14 @@ do a docker build locally but inside of the development version of CCF image:
 - Run a build inside of the CCF docker image:
 
     ```
+    CCF_VERSION="3.0.12"
     docker run -it --rm \
         -w /__w/1/s -v $(pwd):/__w/1/s \
         -v /var/run/docker.sock:/var/run/docker.sock \
         --env PLATFORM=sgx \
         --env CXXFLAGS="-ferror-limit=0" \
         --env NINJA_FLAGS="-k 0" \
-        mcr.microsoft.com/ccf/app/dev:3.0.12-sgx git config --global --add safe.directory "*" && ./docker/build.sh
+        mcr.microsoft.com/ccf/app/dev:"$CCF_VERSION"-sgx git config --global --add safe.directory "*" && ./docker/build.sh
     ```
 - The build will print the value of MRENCLAVE in the log, similar to:
 
--- a/docs/reproducibility.md
+++ b/docs/reproducibility.md
@@ -11,18 +11,18 @@ You need a couple pieces of information to begin:
     {
         "endorsements": "AQAAAAIAAADZL...UMTI6NTU6MTFaAA==",
         "format": "OE_SGX_v1",
-        "mrenclave": "b717f0d475e45aa0294f8229ca646316a17dcd7e4ba7f644b353a5d135a6665f",
+        "mrenclave": "fb2c496416fbab20837fedda0ba6db58d819fa5f5c1b3916062eb2fb9d889966",
         "node_id": "247f1df23e22256cc5bc5e8822183117bc5967da41a257d307f9b1153a4f1853",
         "raw": "AwACAAAAAAAIAA...ViynsClboLw="
     }
     ```
-- Git commit id that built this version. This is something that has no specifc mapping at the moment. You would need to check the build logs of the SGX Docker image to understand which commit produced the candidate value.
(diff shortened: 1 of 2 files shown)


docs/reproducibility.md
```diff
@@ -0,0 +1,44 @@
+# Reproducibility
+
+The assumption here is that the original build was done using a Docker build. The
+goal is to reproduce the same MRENCLAVE value. 
+
+You need a couple pieces of information to begin:
+
+- The source MRENCLAVE value, get it from `https://ledgerurl/node/quotes/self`, e.g.:
```
andpiccione, line 8: Probably not relevant, feel free to ignore: this endpoint would return the MRENCLAVE / quotes data for the specific node in the ledger the client would connect to. If there is more than one node in the network, It is possible that other nodes in the same CCF network have a different MRENCLAVE (e.g., during a code upgrade).

Do you think it could make sense to specify this, or warn users they should check the quotes data for every node if they want to validate all the code versions running in the ledger?

andpiccione, line 8: nit: just to make sure users would know they have to replace their ledger name in the given URL

```suggestion
- The source MRENCLAVE value, get it from `https://<LEDGER-URL>/node/quotes/self` (replace <LEDGER-URL> with the URL of your ledger), e.g.:
```

docs/reproducibility.md
```diff
@@ -0,0 +1,44 @@
...
+- Run a build inside of the CCF docker image:
+
+    ```
+    docker run -it --rm \
+        -w /__w/1/s -v $(pwd):/__w/1/s \
+        -v /var/run/docker.sock:/var/run/docker.sock \
+        --env PLATFORM=sgx \
+        --env CXXFLAGS="-ferror-limit=0" \
+        --env NINJA_FLAGS="-k 0" \
+        mcr.microsoft.com/ccf/app/dev:3.0.12-sgx git config --global --add safe.directory "*" && ./docker/build.sh
```
andpiccione, line 35: nit: Just to make it easier to update the CCF version and make the command more generic

```suggestion
    CCF_VERSION="3.0.12"
    docker run -it --rm \
        -w /__w/1/s -v $(pwd):/__w/1/s \
        -v /var/run/docker.sock:/var/run/docker.sock \
        --env PLATFORM=sgx \
        --env CXXFLAGS="-ferror-limit=0" \
        --env NINJA_FLAGS="-k 0" \
        mcr.microsoft.com/ccf/app/dev:"$CCF_VERSION"-sgx git config --global --add safe.directory "*" && ./docker/build.sh
```

docs/reproducibility.md
```diff
@@ -0,0 +1,44 @@
...
+        "raw": "AwACAAAAAAAIAA...ViynsClboLw="
+    }
+    ```
+- Git commit id that built this version. This is something that has no specifc mapping at the moment. You would need to check the build logs of the SGX Docker image to understand which commit produced the candidate value. e.g., `fb2c496416fbab20837fedda0ba6db58d819fa5f5c1b3916062eb2fb9d889966` was built from `fd77c0c69ee890bdc2fcf6ef0c9dddb7b211e164`.
+
+To reproduce the same MRENCLAVE value which would be deployed to CCF 
+do a docker build locally but inside of the development version of CCF image:
+
+- Clone the repository and check out the tag or commit id that built the binary which had specific MRENCLAVE.
+- Identify the expected CCF version by inspecting `./docker/enclave.Dockerfile`.
```
andpiccione, line 25: It is possible to get the CCF version also by checking the `/node/version` endpoint on a live instance. Do you think it's worth specifying this here as an alternative?

andpiccione, line 25: small nit: worth referencing the dockerfile inside the repo? 

```suggestion
- Identify the expected CCF version by inspecting the [Dockerfile](docker/enclave.Dockerfile).
```

andpiccione (approved):
LGTM, left a few minor comments.

docs/reproducibility.md
ivarprudnikov: I think no because it is not a guarantee that it was used to build the image

docs/reproducibility.md
ivarprudnikov: No, I think it detracts from the goal of the doc - it is about reproducibility, not about CCF and the way nodes get upgraded

PR #156, opened 2023-09-25
Upgrade to CCF 4.0.7

andpiccione:
- Update SCITT app to CCF 4.0.7 and modify source code, tests, and pipelines definition following the new version updates
- The Virtual ASAN build job has been modified to be able to build CCF from source correctly by installing the required dependencies
- Dockerfiles have been modified to revert the changes made in https://github.com/microsoft/scitt-ccf-ledger/pull/144

Note: the app was not updated to the latest CCF patch versions (4.0.8 or 4.0.9) since those include a major version upgrade of OpenSSL that requires more work and introduce a non-minimal performance regression (~20% drop in throughput) for operations that require certificate authentication.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -40,4 +40,4 @@ Checks: >
 
 WarningsAsErrors: '*'
 # Include current directory files and exclude ccf imported code
-HeaderFilterRegex: '^(?!.*\/opt\/ccf_virtual\/include\/).*'
+HeaderFilterRegex: '^(?!.*(\/opt_ccf.*\/include\/)).*'
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -39,4 +39,5 @@ Checks: >
   -readability-qualified-auto,
 
 WarningsAsErrors: '*'
-HeaderFilterRegex: '.*'
+# Include current directory files and exclude ccf imported code
+HeaderFilterRegex: '^(?!.*\/opt\/ccf_virtual\/include\/).*'
--- a/.pipelines/ccf.yml
+++ b/.pipelines/ccf.yml
@@ -11,17 +11,24 @@ parameters:
 
 steps:
   - script: |
+      set -ex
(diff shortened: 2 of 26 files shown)


.clang-tidy
```diff
@@ -39,4 +39,5 @@ Checks: >
   -readability-qualified-auto,
 
 WarningsAsErrors: '*'
-HeaderFilterRegex: '.*'
+# Include current directory files and exclude ccf imported code
+HeaderFilterRegex: '^(?!.*\/opt\/ccf_virtual\/include\/).*'
```
ivarprudnikov, line 43: Please also include `ccf_sgx` as that could be another platform it is built on

.pipelines/ccf.yml
```diff
@@ -11,11 +11,24 @@ parameters:
...
-      cmake -GNinja -B build \
+      set -ex
+      cd getting_started/setup_vm
+      ./run.sh ccf-dev.yml -e ccf_ver=${{ parameters.Version }} -e platform=${{ parameters.Platform }} -e clang_version=15
+    workingDirectory: $(Pipeline.Workspace)/CCF
+    displayName: Setup CCF development environment
+
+  - script: |
+      set -ex
+      sudo rm -rf build
```
ivarprudnikov, line 28: this directory should not exist as it is running on a fresh agent

vimauro: the agent might not be necessarily new, so I'd keep this

vimauro (approved):

.clang-tidy
andpiccione: Made the regex more generic to include any platform.

.pipelines/ccf.yml
andpiccione: I had the same thought. While debugging other issues, I noticed that CCF do the same in their build, so I thought it might not hurt too have it here too just in case.

PR #157, opened 2023-09-28
Replaced hosted pool and Container Registry connection

vimauro:
This PR replaces the hosted pool to be used by Azure Pipelines for internal SGX builds and the service connection used to push images to a staging Azure Container Registry.

### DIFF
--- a/.pipelines/azure-pipelines.yml
+++ b/.pipelines/azure-pipelines.yml
@@ -24,7 +24,7 @@ variables:
   # This is used in functional tests to make certain tests mandatory
   # in CI, but optional during local development. 
   SCITT_CI: 1
-  ACR_REGISTRY: scittoss.azurecr.io
+  ACR_REGISTRY_STAGING: confidentialledgeracrstaging.azurecr.io
   ACR_REGISTRY_RELEASE: confidentialledgeracrprod.azurecr.io
 
 stages:
@@ -143,22 +143,22 @@ stages:
               path: $(Pipeline.Workspace)
 
           - script: |
-              echo "##vso[task.setvariable variable=git_tag]${BUILD_SOURCEBRANCH#refs/tags/}"
+              echo "##vso[task.setvariable variable=git_tag]${BUILD_SOURCEBRANCH#refs/heads/main/}"
             displayName: Get image tag from git tag
           
           - script: |
               docker load --input $(Pipeline.Workspace)/scitt-sgx.tar
-              docker tag scitt-sgx $(ACR_REGISTRY)/scitt-sgx:$(git_tag)
+              docker tag scitt-sgx $(ACR_REGISTRY_STAGING)/scitt-sgx:$(git_tag)
             displayName: Load Docker image
 
           - task: Docker@2
             displayName: Login to ACR
             inputs:
               command: login
-              containerRegistry: scittoss-acr-conn
+              containerRegistry: scitt-acr-conn-staging
 
           - script: |
-              docker push $(ACR_REGISTRY)/scitt-sgx:$(git_tag)
+              docker push $(ACR_REGISTRY_STAGING)/scitt-sgx:$(git_tag)
             displayName: Publish to ACR
 
   - ${{ if startsWith(variables['Build.SourceBranch'], 'refs/tags/') }}:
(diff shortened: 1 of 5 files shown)


.pipelines/azure-pipelines.yml
```diff
@@ -24,7 +24,6 @@ variables:
   # This is used in functional tests to make certain tests mandatory
   # in CI, but optional during local development. 
   SCITT_CI: 1
-
   ACR_REGISTRY: scittoss.azurecr.io
```
andpiccione, line 27: Any reason to maintain a different registry and not push to the existing `confidentialledgeracrstaging` ACR? Just trying to understand if it's worth maintaining a different ACR at this point if we are not planning to onboard to MAR.

andpiccione (approved):
lgtm, left some minor comments

.pipelines/azure-pipelines.yml
```diff
@@ -148,15 +147,15 @@ stages:
           
           - script: |
               docker load --input $(Pipeline.Workspace)/scitt-ccf-ledger-sgx.tar
-              docker tag scitt-ccf-ledger-sgx $(ACR_REGISTRY)/public/scitt-ccf-ledger/app/run:$(git_tag)-sgx
+              docker tag scitt-ccf-ledger-sgx $(ACR_REGISTRY)/public/scitt-ccf-ledger:$(git_tag)-sgx
```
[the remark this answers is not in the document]
vimauro: Sounds good.
Currently we are publishing only SGX images, do you suggest publishing also the `virtual `ones?

.pipelines/azure-pipelines.yml
vimauro: My preference is to separate SCITT entirely to be honest. It's easy to replace it any time in future if needed, the biggest issue was configuring the pools and Az pipelines on ADO

.pipelines/azure-pipelines.yml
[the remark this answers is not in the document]
andpiccione: No I think it's fine to publish only SGX ones for now. We don't need to use the `virtual` ones for testing, so it's not needed imo.

.pipelines/azure-pipelines.yml
andpiccione: Sounds good. However, keep in mind this means we need to configure required permissions for our AKS clusters to pull images from the new registry and modify the scitt helm chart accordingly to pull the CCF app container image from the new registry.

andpiccione (approved):

.pipelines/azure-pipelines.yml
vimauro: This is a valid point but let's discuss more offline, I think that depends on some ongoing discussion about the distribution of the service

.pipelines/azure-pipelines.yml
```diff
@@ -24,8 +24,8 @@ variables:
   # This is used in functional tests to make certain tests mandatory
   # in CI, but optional during local development. 
   SCITT_CI: 1
-
   ACR_REGISTRY: scittoss.azurecr.io
+  ACR_REGISTRY_RELEASE: confidentialledgeracrprod.azurecr.io
```
andpiccione, line 28: If we are pushing release images to the existing ACL prod ACR, why can't we also push the main-branch images to the ACL staging ACR instead of maintaining a different registry?

As said above, pushing and pulling to a new SCITT ACR would require some minor changes and permissions setup. I'm fine if we prefer to use it, but I believe we should either always use ACL ACRs (in both staging and prod) or have separate SCITT ACRs (in both staging and prod).

Sorry, I know I'm a broken record here, but I just want to make sure we commit to either one or the other: if we start mixing ACL and SCITT ACR, it may get confusing.

andpiccione, line 28: Do you think it could make sense to push release images to both staging and prod ACRs, so that we can test new releases in staging too? If we only have main-branch images, we cannot use the images specifically created for new SCITT releases

vimauro: I am more incline on creating a new ACR for scitt prod but as we discussed, it can be quite challenging as of today. 
I will switch from `scittoss `to `confidentialledgerstaging `for now

vimauro: yup makes sense, I just made the change and created a new yml template to reduce duplication

vimauro: Agreed.
I just created a new service connection and switched the staging registry

andpiccione (approved):

PR #160, opened 2023-10-24
defaults to did resolver if service params not passed when validating

ivarprudnikov:
Currently ledger can issue receipts with DID but the CLI cannot validate them using the DID resolver. The CLI expects service params to be present.

Adding a fallback to use the DID resolver if the service params are not provided when validating the receipt

### DIFF
--- a/pyscitt/pyscitt/cli/validate_cose.py
+++ b/pyscitt/pyscitt/cli/validate_cose.py
@@ -4,7 +4,7 @@ import argparse
 from pathlib import Path
 from typing import Optional
 
-from ..verify import StaticTrustStore, DIDResolverTrustStore, verify_receipt
+from ..verify import DIDResolverTrustStore, StaticTrustStore, TrustStore, verify_receipt
 
 
 def validate_cose_with_receipt(
@@ -19,6 +19,7 @@ def validate_cose_with_receipt(
     else:
         receipt = receipt_path.read_bytes()
 
+    service_trust_store: TrustStore
     if service_trust_store_path is None:
         service_trust_store = DIDResolverTrustStore()
     else:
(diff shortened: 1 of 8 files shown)


pyscitt/pyscitt/cli/validate_cose.py
```diff
@@ -4,21 +4,27 @@
...
-
     cose = cose_path.read_bytes()
 
     if receipt_path is None:
         receipt = None
     else:
         receipt = receipt_path.read_bytes()
 
+    service_trust_store: TrustStore
+    if service_trust_store_path is None:
```
andpiccione, line 23:
```suggestion
    if not service_trust_store_path:
```

pyscitt/pyscitt/prefix_tree.py
```diff
@@ -104,7 +104,7 @@ def from_cose_obj(cls, cose_obj: list) -> "ReadReceipt":
 
     @classmethod
     def decode(cls, data: bytes):
-        return cls.from_cose_obj(cbor2.loads(data))
+        return cls.from_cose_obj(cbor2.loads(data))  # type: ignore[arg-type]
```
andpiccione, line 107: Just for my understanding, why did we add all these type check ignores in other files?

andpiccione (approved):

pyscitt/pyscitt/prefix_tree.py
ivarprudnikov: because the checks started to fail out of the blue, I believe this is to do with the unpinned dependencies

ivarprudnikov: What i am trying to say is that nothing changed in the logic so I just ignored these as it seemed like noise

PR #169, opened 2023-12-14
Add GitHub action to publish pyscitt CLI to PyPi

andpiccione:
- Add simple GitHub action to publish `pyscitt` python CLI to PyPi whenever a new SCITT release is published
- Refactor `pyscitt` setup file to include more details useful for packaging purposes

Tested the action against the TestPyPi index: https://test.pypi.org/project/pyscitt/

image


pyscitt/README.md
```diff
@@ -1,18 +1,7 @@
...
 
-The CLI is extensively used in functional tests and in demo scripts:
+Tools to sign claims and interact with a SCITT CCF Ledger.
 
-- [Transparency service demo](../demo/cts_poc/README.md)
-- [GitHub hosted DID demo](../demo/github/README.md)
-- [CLI tests](../test/test_cli.py)
+For more information, please find the `scitt-ccf-ledger` repository at https://github.com/microsoft/scitt-ccf-ledger.
 
-## Installation
```
ivarprudnikov, line 10: The instructions have to go somewhere, otherwise it will be confusing to anyone who lands in this repository

.github/workflows/publish-pyscitt-to-pypi.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          python-version: '3.8'
+
+      - name: Build package
+        run: |
+          python3 -m pip install build
+          cd pyscitt
+          python3 -m build
+      
+      - name: Publish package to PyPI
```
ivarprudnikov, line 36: Instead of having it as a separate action just merge it to `publish-gh-release-files.yml` and maybe rename it to reflect that it is an action that does something on tag pushes.

It is basically a duplicate but instead of storing the built files in a github release it pushes it to pypi

ivarprudnikov (changes requested):

pyscitt/README.md
andpiccione: Well, I removed these instructions because they do not make much sense once we merge this PR: they just provide steps on how to install pyscitt from the GH release, but once we publish the CLI to PyPi, users will be able to just install the package directly without the need to follow these steps.

In any case, I moved these instructions to the README for the moment (since the `pyscitt/README.md` will be used for the long package description available in PyPi) and we can refactor them at a later point.

.github/workflows/publish-pyscitt-to-pypi.yml
andpiccione: I created a separate action to make the two jobs run independently and because I didn't want to modify the existing job to create the GitHub release. But indeed they are sharing most of the steps and it can make sense to merge them.

I created a single `release` action with two independent jobs to create a GH release and publish pyscitt to PyPi. Let me know what you think.

ivarprudnikov (approved):

PR #176, opened 2024-01-15
adds mrenclave to receipt header

ivarprudnikov:
To aid better visibility into the state of the source code that was used to build the scitt enclave, `mrenclave` measurement was added to the receipt headers. Should increase the root of trust of this service.

Contrary to just the DID or service identity this ensures one can look up the source code and to find and rebuild the same version locally. If does not solve the problem how to find the correct source code version though, that could be solved by printing the value in the releases on GitHub, but for now mrenclave is printed in the CI build.


test/test_x509.py
```diff
@@ -44,6 +45,23 @@ def test_submit_claim_x5c(
...
+    expected_virtual_measurement = (
+        "0000000000000000000000000000000000000000000000000000000000000000"
+    )
+    if env_platform == "virtual":
+        assert actual_measurement == expected_virtual_measurement
+    elif env_platform == "sgx":
+        assert (
+            len(actual_measurement) == 64
+            and actual_measurement != expected_virtual_measurement
+        )
```
andpiccione, line 61: A simpler way here would be to get the `mrenclave` from the running CCF network using the `/node/quotes/self` endpoint and compare that this matches with the mrenclave included in the receipt.

app/src/main.cpp
```diff
@@ -219,6 +219,25 @@ namespace scitt
...
+      {
+        measurement = node_info->code_digest.value();
+      }
+      else
+      {
+#if defined(INSIDE_ENCLAVE) && !defined(VIRTUAL_ENCLAVE)
+        // Node should always get a valid cached measurement on startup
+        throw InternalError("Unexpected state - node has no code id");
+#else
+        measurement = "";
```
andpiccione, line 237: Why set it to an empty string if it could not be retrieved when running in virtual platform? IIUC, it would result in a different mrenclave value in the receipt, as it's not the same as a 64-0s string, making it potentially inconsistent.

test/test_x509.py
ivarprudnikov: Yes and no, because in a test you want to understand what value is expected.

app/src/main.cpp
ivarprudnikov: updated to the expected one

andpiccione (approved):

PR #196, opened 2024-07-13
Fix CBOR encoder buffer overflow for large claims

andpiccione:
Fetching a large claim with an embedded receipt can result in an error if the size of the claim + the size of the receipt payload exceeds the size allocated to the CBOR encoder buffer (default to 10KB).

This PR fixes the issue by initializing a dynamic buffer size proportionally to the size of the claim and the receipt when embedding a receipt into a claim entry. Other relevant changes included:

- Added a functional test to reproduce the bug and validate the fix.
- Added a flag in the `pyscitt retrieve_signed_claims` command to allow users to optionally embed receipts in claims.
- Added more logging to make the debugging and investigation of similar problems easier in the future.

### DIFF
--- a/pyscitt/pyscitt/cli/retrieve_signed_claims.py
+++ b/pyscitt/pyscitt/cli/retrieve_signed_claims.py
@@ -16,6 +16,7 @@ def retrieve_signed_claimsets(
     from_seqno: Optional[int],
     to_seqno: Optional[int],
     service_trust_store_path: Optional[Path],
+    dont_embed_receipt: bool = False,
 ):
     base_path.mkdir(parents=True, exist_ok=True)
 
@@ -25,7 +26,7 @@ def retrieve_signed_claimsets(
         service_trust_store = None
 
     for tx in client.enumerate_claims(start=from_seqno, end=to_seqno):
-        claim = client.get_claim(tx, embed_receipt=True)
+        claim = client.get_claim(tx, embed_receipt=(not dont_embed_receipt))
         path = base_path / f"{tx}.cose"
 
         if service_trust_store:
@@ -52,11 +53,22 @@ def cli(fn):
         type=Path,
         help="Folder containing JSON parameter files of SCITT services to trust",
     )
+    parser.add_argument(
+        "--dont-embed-receipt",
+        action="store_true",
+        default=False,
+        help="Whether to not return the signed claimsets with the receipt embedded. Default is False.",
+    )
 
     def cmd(args):
         client = create_client(args)
         retrieve_signed_claimsets(
-            client, args.path, args.from_seqno, args.to_seqno, args.service_trust_store
+            client,
+            args.path,
+            args.from_seqno,
+            args.to_seqno,
+            args.service_trust_store,
+            args.dont_embed_receipt,
         )
 
     parser.set_defaults(func=cmd)
--- a/pyscitt/pyscitt/cli/retrieve_signed_claims.py
+++ b/pyscitt/pyscitt/cli/retrieve_signed_claims.py
@@ -16,7 +16,7 @@ def retrieve_signed_claimsets(
     from_seqno: Optional[int],
     to_seqno: Optional[int],
     service_trust_store_path: Optional[Path],
-    dont_embed_receipt: bool = False,
+    embed_receipt: bool = False,
 ):
     base_path.mkdir(parents=True, exist_ok=True)
 
@@ -26,7 +26,7 @@ def retrieve_signed_claimsets(
         service_trust_store = None
 
     for tx in client.enumerate_claims(start=from_seqno, end=to_seqno):
-        claim = client.get_claim(tx, embed_receipt=(not dont_embed_receipt))
+        claim = client.get_claim(tx, embed_receipt=embed_receipt)
         path = base_path / f"{tx}.cose"
 
         if service_trust_store:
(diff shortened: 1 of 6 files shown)


pyscitt/pyscitt/cli/retrieve_signed_claims.py
```diff
@@ -25,7 +26,7 @@ def retrieve_signed_claimsets(
         service_trust_store = None
 
     for tx in client.enumerate_claims(start=from_seqno, end=to_seqno):
-        claim = client.get_claim(tx, embed_receipt=True)
+        claim = client.get_claim(tx, embed_receipt=(not dont_embed_receipt))
```
ivarprudnikov, line 29:
```suggestion
        claim = client.get_claim(tx, embed_receipt=include_receipt)
```
Avoiding double negative

pyscitt/pyscitt/cli/retrieve_signed_claims.py
```diff
@@ -52,11 +53,22 @@ def cli(fn):
         type=Path,
         help="Folder containing JSON parameter files of SCITT services to trust",
     )
+    parser.add_argument(
+        "--dont-embed-receipt",
```
ivarprudnikov, line 57:
```suggestion
        "--include-receipt",
```
Avoiding double negative

pyscitt/pyscitt/cli/retrieve_signed_claims.py
```diff
@@ -52,11 +53,22 @@ def cli(fn):
...
+        action="store_true",
+        default=False,
```
ivarprudnikov, line 59:
```suggestion
        default=True,
```

pyscitt/pyscitt/cli/retrieve_signed_claims.py
```diff
@@ -52,11 +53,22 @@ def cli(fn):
...
+        help="Whether to not return the signed claimsets with the receipt embedded. Default is False.",
```
ivarprudnikov, line 60:
```suggestion
        help="Whether to include the receipt in the signed claimset under the unprotected header. Default is True.",
```
Avoiding double negative as the text otherwise is quite confusing to the people new to this

pyscitt/pyscitt/cli/retrieve_signed_claims.py
```diff
@@ -52,11 +53,22 @@ def cli(fn):
...
     def cmd(args):
         client = create_client(args)
         retrieve_signed_claimsets(
-            client, args.path, args.from_seqno, args.to_seqno, args.service_trust_store
+            client,
+            args.path,
+            args.from_seqno,
+            args.to_seqno,
+            args.service_trust_store,
+            args.dont_embed_receipt,
```
ivarprudnikov, line 71:
```suggestion
            args.include_receipt,
```

pyscitt/pyscitt/cli/retrieve_signed_claims.py
```diff
@@ -16,6 +16,7 @@ def retrieve_signed_claimsets(
     from_seqno: Optional[int],
     to_seqno: Optional[int],
     service_trust_store_path: Optional[Path],
+    dont_embed_receipt: bool = False,
```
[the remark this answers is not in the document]
andpiccione: I explicitly used a double negative here to minimize breaking changes. By default, the current CLI retrieves all claims with embedded receipts. Introducing an optional flag for embedding receipts would mean that clients would start getting claims without embedded receipts by default, unless they use the new flag.

While considering this change, I preferred using a double negative to avoid breaking changes; however, if we are not concerned about it and prefer to make the CLI arguments clearer, I can definitely make the change to avoid the double negative.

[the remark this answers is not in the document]
ivarprudnikov: I would not be concerned with such a breaking change because getting the signature back without a receipt does not make much sense. I would consider it a feature because prior to that you needed to explicitly set the argument to get the embedded sig.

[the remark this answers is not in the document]
andpiccione: Fair enough. Addressed in https://github.com/microsoft/scitt-ccf-ledger/pull/196/commits/d22216a3994850124c91db08b959fd06bb8aae73

pyscitt/pyscitt/cli/retrieve_signed_claims.py
andpiccione: See https://github.com/microsoft/scitt-ccf-ledger/pull/196#discussion_r1678073658

pyscitt/pyscitt/cli/retrieve_signed_claims.py
andpiccione: See https://github.com/microsoft/scitt-ccf-ledger/pull/196#discussion_r1678073658

pyscitt/pyscitt/cli/retrieve_signed_claims.py
andpiccione: It doesn't make sense to set the default value for this flag to `True`: how can users set it to `False` if they pass the flag? By default, it will be False and adding the CLI argument will set it to True.

pyscitt/pyscitt/cli/retrieve_signed_claims.py
andpiccione: See https://github.com/microsoft/scitt-ccf-ledger/pull/196#discussion_r1678073658

ivarprudnikov (approved):

PR #202, opened 2024-07-23
Testing changes to compare with OB

maxtropets:


test/infra/cchost.py
```diff
@@ -182,27 +182,22 @@ async def _start_process(self) -> None:
...
+        stderr_file = open(f"{self.workspace}/std.err", "w")
         process = await asyncio.create_subprocess_exec(
             self.binary,
             "--config",
             self.workspace / "config.json",
             cwd=self.workspace,
             start_new_session=True,
             stdin=subprocess.DEVNULL,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
```
achamayou, line 193: see https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait

> Note This will deadlock when using stdout=PIPE or stderr=PIPE and the child process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use [Popen.communicate()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) when using pipes to avoid that.

achamayou (approved):

test/infra/cchost.py
maxtropets: You mean that could be an issue before? Or is it a suggestion to some change?

test/infra/cchost.py
```diff
@@ -182,27 +182,22 @@ async def _start_process(self) -> None:
             cchost_env["FAKETIME_NO_CACHE"] = "1"
 
         LOG.debug("Starting cchost process...")
+        stdout_file = open(f"{self.workspace}/std.out", "w")
+        stderr_file = open(f"{self.workspace}/std.err", "w")
```
andpiccione, line 186: Could it make sense to print the content of these files when the cchost process terminates to improve the debugging of possible cchost errors? If we just write them into files, it's hardly useful unless someone debugs the tests locally and check for these files explicitly

andpiccione, line 186: Wouldn't these files be overwritten if cchost is restarted by the tests during the same pytest run? It could make sense to write these files in unique places (e.g., with a timestamp or test ID in the file name) to ensure that the captured logs can be referenced to a specific test.

.github/workflows/build-test-virtual.yml
```diff
@@ -5,6 +5,10 @@ on:
       - main
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
andpiccione, line 10: Why do we want to ensure that a single job run at a time?

andpiccione (approved):

test/infra/cchost.py
achamayou: The idea is to upload them for later inspection, with upload-artifact in GHA or equivalent. Then you can download them all if you need to debug something.

achamayou: Yes, CCF tests create per-test and per-node directories for that exact reason, that's definitely something SCITT tests should do too.

.github/workflows/build-test-virtual.yml
achamayou: This is a single job per-branch at a time. So if you push commit A and later commit B, the build for commit A gets stopped at the one for commit B gets started. It's ok to continue running old builds, but it's wasteful, because even if it passes, it doesn't tell you if the build for commit B will.

.github/workflows/build-test-virtual.yml
```diff
@@ -36,4 +40,12 @@ jobs:
...
       - run: ./build.sh
       - run: ./run_unit_tests.sh
-      - run: ./run_functional_tests.sh --enable-prefix-tree --enable-faketime
+      - run: ./run_functional_tests.sh --enable-faketime
+
+      - name: Upload logs
+        uses: actions/upload-artifact@v4
+        if: ${{ failure() }}
+        with:
+          name: out
```
achamayou, line 49: Is that right? Is self.workspace really out/? Even so, it makes sense to only upload the logs, the binaries will be large, and the upload will take a long time. See the CCF ci for a sample.

test/infra/cchost.py
maxtropets: There are being uploaded afterwards

maxtropets: Same in GH pipelines, but on failure only

.github/workflows/build-test-virtual.yml
maxtropets: this

.github/workflows/build-test-virtual.yml
maxtropets: It's only logs now, no one else uses it.

The idea of OB pipeline is that this's the directory you want to save after the run, so you only write there something you might need to investigate the failure.

test/infra/cchost.py
maxtropets: It's controlled by the fixtures. IIUC, currently all tests except marked as @isolated_test reuse the same setup, and so they share the log.

Isolated have their own `workspace` paths, so they don't interfere.

maxtropets: You might want to change that and have more isolated tests, but I think it out of the scope of this PR.

PR #203, opened 2024-07-24
Add configurable policy via sandboxed JS execution

eddyashton:
Demonstrating how the JS interpreter API added in CCF 5 could be used to implement dynamic/customisable policy.

In short:

- The (governance-populated) `Configuration::Policy` struct now has an optional `policy_script` field
- If this is present, we execute it as a JS script:
  - look for an `apply()` function
  - pass the protected headers extracted from the COSE envelope
  - expect it to return a bool indicating whether submission should be accepted
- If the policy exists and fails, return an error

Includes some tests of the error paths, and of a simple sample policy which protects some feeds so they can only be claimed by a given issuer. These feeds and issuers are currently literals within the policy script, but they could also be read from the KV.

### DIFF
--- a/app/constitution/scitt.js
+++ b/app/constitution/scitt.js
@@ -24,6 +24,8 @@ actions.set("set_scitt_configuration",
             checkType(alg, "string", `configuration.policy.accepted_did_issuers[${i}]`);
           }
         }
+
+        // TODO: Validate type of policy_script
       }
 
       checkType(args.configuration.authentication, "object?", "configuration.authentication");
--- a/app/constitution/scitt.js
+++ b/app/constitution/scitt.js
@@ -24,8 +24,7 @@ actions.set("set_scitt_configuration",
             checkType(alg, "string", `configuration.policy.accepted_did_issuers[${i}]`);
           }
         }
-
-        // TODO: Validate type of policy_script
+        checkType(args.configuration.policy.policy_script, "string?", "configuration.policy.policy_script");
       }
 
       checkType(args.configuration.authentication, "object?", "configuration.authentication");
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -253,14 +253,24 @@ namespace scitt
 
       if (cfg.policy.policy_script.has_value())
       {
-        if (!scitt::run_policy_engine(
-              cfg.policy.policy_script.value(),
-              "configured_policy",
-              claim_profile,
-              phdr))
+        SCITT_DEBUG("Applying policy");
+        const auto policy_violation_reason = check_for_policy_violations(
+          cfg.policy.policy_script.value(),
+          "configured_policy",
+          claim_profile,
+          phdr);
+        if (policy_violation_reason.has_value())
         {
-          throw BadRequestError(errors::PolicyFailed, "Policy was not met");
+          throw BadRequestError(
+            errors::PolicyFailed,
+            fmt::format(
+              "Policy was not met: {}", policy_violation_reason.value()));
         }
+        SCITT_DEBUG("Policy check passed");
+      }
+      else
+      {
+        SCITT_DEBUG("No policy applied");
       }
 
       // TODO: Apply further acceptance policies.
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -33,7 +33,7 @@
 #include <ccf/historical_queries_interface.h>
 #include <ccf/http_query.h>
 #include <ccf/indexing/strategies/seqnos_by_key_bucketed.h>
-#include <ccf/js/core/context.h>
+#include <ccf/js/common_context.h>
 #include <ccf/js/extensions/console.h>
 #include <ccf/json_handler.h>
 #include <ccf/kv/value.h>
@@ -61,6 +61,13 @@ const x = 2;
 const y = 3;
 console.log(`x = ${x}, y = ${y}`);
 console.log(`x+y = ${x+y}`);
+
+export function apply(s) {
+  console.log(`Calling apply function`);
+  console.log(`Argument is: ${s}`);
+  // Return reversed string
+  return s.split('').reverse().join('');
+}
 )!!!";
 
 namespace scitt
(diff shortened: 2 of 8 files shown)


app/src/policy_engine.h
```diff
@@ -0,0 +1,214 @@
...
+            reason,
+            trace.value_or("<no trace>")));
+      }
+
+      // JS-style semantics - anything truthy becomes true
+      return result.is_true();
+    }
+  }
+
+  static inline bool run_policy_engine(
```
andpiccione, line 206: Could it make sense to also pass the claim payload or the unprotected headers for cases where the policy should be applied there too?

app/src/main.cpp
```diff
@@ -248,6 +251,18 @@ namespace scitt
 #endif
       }
 
+      if (cfg.policy.policy_script.has_value())
```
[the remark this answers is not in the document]
eddyashton: Good point! I've now added logging in each branch to distinguish the 3 cases (no policy, policy failed, policy accepted).

app/src/main.cpp
```diff
@@ -248,6 +251,18 @@ namespace scitt
...
+      {
+        if (!scitt::run_policy_engine(
+              cfg.policy.policy_script.value(),
+              "configured_policy",
+              claim_profile,
+              phdr))
+        {
+          throw BadRequestError(errors::PolicyFailed, "Policy was not met");
```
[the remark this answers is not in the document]
eddyashton: Sure, I've updated the JS wrapping code to support that. The current implementation is variadic return type from the JS function ("true-or-error-string"), but if anyone has a cleaner suggestion we could do that too. Might be sensible to just require some structured object already, so we have some extensibility in there.

app/constitution/scitt.js
```diff
@@ -24,6 +24,7 @@ actions.set("set_scitt_configuration",
             checkType(alg, "string", `configuration.policy.accepted_did_issuers[${i}]`);
           }
         }
+        checkType(args.configuration.policy.policy_script, "string?", "configuration.policy.policy_script");
```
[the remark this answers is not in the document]
eddyashton: Thanks for the pointer - docs added.

app/src/policy_engine.h
achamayou: +1 for not passing the unprotected headers, they are by definition unsuitable for policy. The only thing I think could make sense is to pass a boolean flag indicating the presence of unprotected headers, so the policy can forbid that. My sense is that this should be forbidden by default though.

Payloads are authenticated, but could be large and CTS is meant to be agnostic to them. Inspection/certification if it takes place should be done by a separate confidential system, which attaches an attestation/policy receipt in protected headers, and CTS should verify those.

Payload inspection can be very expensive in terms of runtime, and large in terms of TCB footprint. It is best if it stays isolated for these reasons. CTS should focus on transparency.

app/src/policy_engine.h
```diff
@@ -0,0 +1,232 @@
...
+  {
+    static inline ccf::js::core::JSWrappedValue claim_profile_to_js_val(
+      ccf::js::core::Context& ctx, ClaimProfile claim_profile)
+    {
+      switch (claim_profile)
+      {
+        case ClaimProfile::IETF:
+        {
+          return ctx.new_string("IETF");
+          break;
```
maxtropets, line 28: return + break?..

eddyashton: Sure! It's a bit of unreachable code that a cautious linter might complain about, but I think it's a sensible pattern - if you replace the `return`s with some local assignment, you're reminded that you're inside a switch statement.

app/src/policy_engine.h
```diff
@@ -0,0 +1,232 @@
...
+          fmt::format("Invalid policy module: {}", e.what()));
+      }
+
+      auto profile_val = claim_profile_to_js_val(interpreter, claim_profile);
+      auto phdr_val = protected_headers_to_js_val(interpreter, phdr);
+
+      const auto result = interpreter.call_with_rt_options(
+        apply_func,
+        {profile_val, phdr_val},
+        std::nullopt,
```
achamayou, line 186:
```suggestion
        std::nullopt, // TODO: add runtime limits (heap, stack, time)
```

test/test_configuration.py
```diff
@@ -110,6 +111,207 @@ def test_multiple_accepted_issuers(
...
+
+        policy_script = f"""
+export function apply(profile, phdr) {{
+    // Only accept x509 submissions with a feed
+    if (profile !== "X509") {{ return "{claim_profile_error}"; }}
+    if (!("feed" in phdr)) {{ return "{missing_feed_error}"; }}
+
+    // Protect access to the first feed
+    // Note this is doing direct cert comparison for simplicity, should
+    // really be basedon a stable issuer ID
```
achamayou, line 239:
```suggestion
    // really be based on a stable issuer ID
```

achamayou (approved):

app/src/policy_engine.h
maxtropets: It just seems too much to me, such as `if (p == nullptr)` instead of `if (!p`, or `5 == x` instead of `x == 5` to prevent accidental assignment.

For reference: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es87-dont-add-redundant--or--to-conditions.

But I'm fine to leave it here, just saying.

achamayou: @eddyashton we could -Wimplicit-fallthrough if we're worried about this, since we have -Werror?

eddyashton: There's a few instances in the code base of returns inside switch statements without a following break (`get_jose_alg_from_cose_alg`, `OperationsIndexingStrategy::lookup`), so I've removed the `break` to match local style.

FWIW, I don't like the core guideline's brevity here - I think `if (p)` for pointers leads to `if (x)` for lots of other expressions with implicit bool conversions, leading to missed errors when the type changes (or for odd implementations of bool operator). eg, if I change a `T x` to `std::optional<T> x`, I'd rather get a compile-error on every existing `if (x.is_good())` (changing to `if (x.has_value())` or `if (x->is_good())` as appropriate), rather than having a bunch of `if (x)` that continue to compile but mean something different.

Anyway this is a very pretty bike shed we've built.

maxtropets: Rust...

app/src/policy_engine.h
```diff
@@ -0,0 +1,229 @@
...
+        fmt::format(
+          "Unexpected return value from policy: {}",
+          interpreter.to_str(result)));
+    }
+  }
+
+  // Returns nullopt for success, else a string describing why the policy was
+  // refused. May also throw if given invalid policies, or policy execution
+  // throws.
+  static inline std::optional<std::string> check_for_policy_violations(
```
maxtropets, line 221: A little counterintuitive that it's inverted by default (usually you're looking for success). But OK.

maxtropets (approved):

andpiccione (approved):

maxtropets (approved):

app/src/policy_engine.h
```diff
@@ -0,0 +1,237 @@
...
+        if (phdr.notary_expiry.has_value())
+        {
+          obj.set_int64("notary_expiry", phdr.notary_expiry.value());
+        }
+      }
+
+      return obj;
+    }
+
+    static inline std::optional<std::string> apply_js_policy(
```
ivarprudnikov, line 154: Please add unit tests to make sure we have coverage of the edge cases, see examples: 
- https://github.com/microsoft/scitt-ccf-ledger/blob/main/app/unit-tests/configurable_auth_test.cpp
- https://github.com/microsoft/scitt-ccf-ledger/blob/main/app/unit-tests/verifier_test.cpp

ivarprudnikov (changes requested):

docs/configuration.md
```diff
@@ -112,20 +113,30 @@ Example `set_scitt_configuration` snippet:
 }
 ```
 
+### Policy script
+JS code that determines whether an entry should be accepted. Should export an `apply` function taking 2 arguments `(claim_profile, protected_headers)`, and return true if the entry should be accepted or a string describing why the entry has failed the policy
```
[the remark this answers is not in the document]
achamayou: done

app/src/policy_engine.h
achamayou: As discussed, added tests for:

1. Invalid JS in body of policy
2. Timeout/infinite loop
3. Heap and Stack limits

ivarprudnikov (approved):

PR #205, opened 2024-08-06
Allow versions higher than python3.8 to run pyscitt 

pavankad:
This precludes the need to install python3.8 when running contract service. Versions above 3.8 are permitted, even the set up file for pyscitt requires >-puython3.8. This saves some time to insall python3.8 when most of the current linux machines run higher version


pavankad:
wrong pull request, only want to commit into kapilvgit:main

PR #206, opened 2024-08-06
did:x509 issuer support in IETF profile

achamayou:
Extension of the changes in #203 to authenticate did:x509 issuers, and updated sample illustrating the policy we expect to see in cases where they are used.

I chose to do this under the IETF profile, rather than x509, but the verification sequence is different from both existing profiles (IETF is only did:web at the moment, as far as I can tell):

1. Find signing key (`phdr.x5chain[0]` currently, but eventually `uhdr.x5chain[0] if digest(uhdr.x5chain[0]) == phdr.x5t`)
2. Verify COSE signature
3. Check issuer is `did:x509`, resolve it against `phdr.x5chain[0]`
4. Check resolved key matches `phdr.x5chain[0]`

### DIFF
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -253,14 +253,24 @@ namespace scitt
 
       if (cfg.policy.policy_script.has_value())
       {
-        if (!scitt::run_policy_engine(
-              cfg.policy.policy_script.value(),
-              "configured_policy",
-              claim_profile,
-              phdr))
+        SCITT_DEBUG("Applying policy");
+        const auto policy_violation_reason = check_for_policy_violations(
+          cfg.policy.policy_script.value(),
+          "configured_policy",
+          claim_profile,
+          phdr);
+        if (policy_violation_reason.has_value())
         {
-          throw BadRequestError(errors::PolicyFailed, "Policy was not met");
+          throw BadRequestError(
+            errors::PolicyFailed,
+            fmt::format(
+              "Policy was not met: {}", policy_violation_reason.value()));
         }
+        SCITT_DEBUG("Policy check passed");
+      }
+      else
+      {
+        SCITT_DEBUG("No policy applied");
       }
 
       // TODO: Apply further acceptance policies.
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -401,9 +401,7 @@ namespace scitt
       SCITT_DEBUG("Register DID:web resolver");
       auto resolver = std::make_unique<did::UniversalResolver>();
       resolver->register_resolver(std::make_unique<did::web::DidWebResolver>());
-
-      auto mock_x509_resolver = std::make_unique<MockMethodResolver>("did:x509:");
-      resolver->register_resolver(mock_x509_resolver);
+      resolver->register_resolver(std::make_unique<did::x509::DidX509Resolver>());
 
       verifier = std::make_unique<verifier::Verifier>(std::move(resolver));
 
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -33,7 +33,7 @@
 #include <ccf/historical_queries_interface.h>
 #include <ccf/http_query.h>
 #include <ccf/indexing/strategies/seqnos_by_key_bucketed.h>
-#include <ccf/js/core/context.h>
+#include <ccf/js/common_context.h>
 #include <ccf/js/extensions/console.h>
 #include <ccf/json_handler.h>
 #include <ccf/kv/value.h>
@@ -61,6 +61,13 @@ const x = 2;
 const y = 3;
 console.log(`x = ${x}, y = ${y}`);
 console.log(`x+y = ${x+y}`);
+
+export function apply(s) {
+  console.log(`Calling apply function`);
+  console.log(`Argument is: ${s}`);
+  // Return reversed string
+  return s.split('').reverse().join('');
+}
 )!!!";
 
 namespace scitt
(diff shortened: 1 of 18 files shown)


app/src/main.cpp
```diff
@@ -248,6 +251,29 @@ namespace scitt
...
+              "Policy was not met: {}", policy_violation_reason.value()));
+        }
+        SCITT_DEBUG("Policy check passed");
+      }
+      else
+      {
+        SCITT_DEBUG("No policy applied");
+      }
+
       // TODO: Apply further acceptance policies.
```
ivarprudnikov, line 277: suppose this comment can be removed now

app/src/verifier.h
```diff
@@ -362,17 +363,106 @@ namespace scitt::verifier
...
-            phdr, tx, current_time, resolution_cache_expiry, configuration);
-
-          try
+          if (phdr.issuer->starts_with("did:x509"))
           {
-            cose::verify(data, key);
+            // IETF SCITT did:x509 claim
+            if (phdr.x5chain.has_value())
+            {
+              // NB: In later revisions of SCITT, x5chain is unprotected, and
```
ivarprudnikov, line 371: Please move this block to a separate method for the consistency sake and to make it a bit more friendly on the eyes, to something like `process_did_x509` or similar.

achamayou: Yes, absolutely.

app/src/main.cpp
achamayou: Yes

app/src/verifier.h
```diff
@@ -362,17 +363,106 @@ namespace scitt::verifier
...
+              // only x5t is. This logic will need to authenticate x5chain[0]
+              // against x5t before it can proceed to verify the signature.
+              // Verify the signature as early as possible
+              OpenSSL::Unique_X509 leaf =
```
ivarprudnikov, line 375: Why is the existing approach not being used: validate chain against the trust store and then use the pub key like in the notary profile processing above, e.g.:

```
auto roots = x509_root_store(tx);
auto cert = verify_chain(roots, x5chain);
key = PublicKey(cert, std::nullopt);
```

pyscitt/pyscitt/crypto.py
```diff
@@ -173,6 +173,7 @@ def generate_cert(
     issuer: Optional[Tuple[Pem, Pem]] = None,
     ca: bool = False,
     cn: Optional[str] = None,
+    inject_eku: Optional[str] = None,
```
ivarprudnikov, line 176: rename form `inject_*` to something more auditor friendly like `add_eku`

app/src/verifier.h
```diff
@@ -362,17 +363,106 @@ namespace scitt::verifier
...
+                pem_chain += ccf::crypto::cert_der_to_pem(c).str();
+              }
+              auto did_document_str = didx509::resolve(
+                pem_chain,
+                phdr.issuer.value(),
+                true /* Do not validate time */);
+              scitt::did::alt::DIDDocument did_document =
+                nlohmann::json::parse(did_document_str);
+
+              if (did_document.verification_method.empty())
```
ivarprudnikov, line 399: Why only verification method is checked. In the did:web case we use only assetion method. I would expect both to be used here for verififcation `assertionMethod` and `verificationMethod`

app/src/verifier.h
```diff
@@ -362,17 +363,106 @@ namespace scitt::verifier
...
+              {
+                throw VerificationError(e.what());
+              }
+              // Then authenticate the did:x509 claim against the x5chain
+              std::string pem_chain;
+              for (auto const& c : phdr.x5chain.value())
+              {
+                pem_chain += ccf::crypto::cert_der_to_pem(c).str();
+              }
+              auto did_document_str = didx509::resolve(
```
ivarprudnikov, line 392: Provided the chain will be verified earlier you could just use `create_did_document` instead. But it would be better to avoid that altogether to remove unnecessary json create/parse. This might otherwise cap the amount of transaction we will be able to process during the heavy load

app/src/verifier.h
achamayou: Because `verify_chain()` is duplicating what `didx509::resolve()` does, but without being apply to apply the query along the way, so we would be checking all the signatures twice effectively.

app/src/verifier.h
achamayou: Because did:x509 does not have assertionMethod? I am not sure what we would check there, can you clarify?

app/src/verifier.h
achamayou: I think what we could do is have didx509::resolve() return a C++ object as well/instead of the JSON document, so we can skip the serialisation/parse. `create_did_document` would not help, because the things `resolve()` ((index, digest) and eku) needs to check are not present in the document and are only in the x5chain.

achamayou: Looked more closely, and I think we can probably simplify the resolve() to a resolve_jwk() that returns a single key, and skip parsing the outer document we don't need here. Let's discuss.

achamayou: As discussed, logged as an optimisation in the did:x509 repo: https://github.com/microsoft/didx509cpp/issues/16

app/src/verifier.h
achamayou: As discussed, this is not necessary because there is only a ever a single key. We can simplify the interface for optimisation purposes by making a change in the library and propagating through CCF to here (logged as https://github.com/microsoft/didx509cpp/issues/16).

ivarprudnikov (approved):
We have identified a couple of possible improvements in the future but as it is the feature is ready to be released the further validated once deployed. Related issues were created to address some challenges.

achamayou:
For reference: #207 #208 and https://github.com/microsoft/didx509cpp/issues/16

PR #211, opened 2024-08-20
Accept CWT_Claims in envelopes, and expose them to policy

achamayou:
- [x] Add test with ESRP payload
- [x] Split CWT_Claims into sub-object for policy
- [x] Add iat support


pyscitt/pyscitt/crypto.py
```diff
@@ -709,6 +709,42 @@ def sign_claimset(
     return msg.encode(tag=True)
 
 
+def sign_claimset_with_cwt(
```
ivarprudnikov, line 712: This should be merged with the above `sign_claimset` purely because that one is used in the CLI to create claims in `sign_claims` method, see https://github.com/microsoft/scitt-ccf-ledger/blob/1415206c7ab085cfbb2c37d37d5d2ce80591152a/pyscitt/pyscitt/cli/sign_claims.py#L155 

That in turn would allow us to write functional tests against the running ledger with the support for creating COSE envelopes with a CWT header.

achamayou: @ivarprudnikov happy to merge stuff, how do you want to expose setting the claims on the CWT?

1. Option to use CWT, but otherwise re-use (issuer, feed, svn)?
2. Specify a CWT object?
3. Just move everything over to CWT?
4. something else?

I also want to add `iat`, as mentioned in the description.

ivarprudnikov: I suppose an option to specify CTW object would follow the existing params and approach to setting headers. Then users will be in full control of what claims to pass there.

achamayou: @ivarprudnikov I think there are two problems with that:

1. There is not an easy way to specify a CWT on the CLI. You'd need to specify the fields you want, the types etc.
2. This allows contradictory issuer/feed/svn information in the final payload.

I propose a `cwt` flag instead, which decides whether the issuer/feed/svn gets set in the protected header, or a CWT_claims map is created with iss/sub/svn/iat. That prevents constructing invalid payloads, and is a much easier switch for users when they get there.

ivarprudnikov: I do not have a strong opinion here. Whatever makes sense for now. This is mainly because we consume this part of the CLI ourselves and this is not really something the end users should use unless for testing.

achamayou:
> One of the small issues remain, but not sure if this needs to be part of this PR. If the detected profile uses CWT and thus becomes IETF profile, the root cert is not checked against the accepted roots, because we expect a policy. But if the policy is not configured then it will just pass through as valid. This means that once the ledger boots up it can start accepting the submissions until the configuration is applied.

All CCF-based services have an explicit [Service Open](https://microsoft.github.io/CCF/main/governance/open_network.html) step, which must be proposed and voted by the Consortium, before user transactions are handled. There is no risk of submissions getting in before the ledger owner configures the policy, other than a deliberate subsequent opening of the service by the owner without policy.

> It might make sense to say that if the profile equals IETF and CWT is used and there is no policy script then reject the submission.

Yes, that's a possibility. Another is to update the constitution to prevent opening altogether if no policy is set. I think there is a broader discussion to be had about keeping both policy and list of CAs, considering the former is more expressive than the latter.

ivarprudnikov (approved):
LGTM

PR #212, opened 2024-08-22
New cli argument allowing to specify receipt type to get after submission

ivarprudnikov:
- New argument `--receipt-type` with options `raw` (default), `embedded`, when calling `submit` action to allow the download of the embedded receipt in one step without going through `embed-receipt` command. This mitigates unwanted side effects when `embed-receipt` logic differs from the server logic; also, this will yield better user experience without the need to explain how to embed receipt in a separate step. It defaults to `raw` to avoid any breaking surprises for those who use the cli already.
- New method `get_last_embedded_receipt_from_cose` which allows extracting the last receipt from a cose header. Used in receipt verification step when only signature with embedded receipt is provided.
- Minor doc improvements

### DIFF
--- a/demo/github/4-submit.sh
+++ b/demo/github/4-submit.sh
@@ -5,12 +5,10 @@
 set -ex
 
 SCITT_URL=${SCITT_URL:-"https://127.0.0.1:8000"}
-SCITT_TRUST_STORE=tmp/trust_store
 
 TMP_DIR=tmp/github
 
 scitt submit $TMP_DIR/claims.cose \
     --receipt $TMP_DIR/claims.receipt.cbor \
     --url "$SCITT_URL" \
-    --service-trust-store $SCITT_TRUST_STORE \
     --development
--- a/pyscitt/pyscitt/cli/submit_signed_claims.py
+++ b/pyscitt/pyscitt/cli/submit_signed_claims.py
@@ -21,9 +21,9 @@ def submit_signed_claimset(
     if path.suffix != ".cose":
         raise ValueError("unsupported file extension, must end with .cose")
 
-    if receipt_type == "raw":
+    if receipt_type == ReceiptType.RAW.value:
         r_type = ReceiptType.RAW
-    elif receipt_type == "embedded":
+    elif receipt_type == ReceiptType.EMBEDDED.value:
         r_type = ReceiptType.EMBEDDED
     else:
         raise ValueError(f"unsupported receipt type {receipt_type}")
@@ -73,8 +73,8 @@ def cli(fn):
     )
     parser.add_argument(
         "--receipt-type",
-        choices=["embedded", "raw"],
-        default="raw",  # default to raw for backwards compatibility
+        choices=[ReceiptType.EMBEDDED.value, ReceiptType.RAW.value],
+        default=ReceiptType.RAW.value,  # default to raw for backwards compatibility
         help="""
         Downloads the receipt of a given type where raw means a countersignature (CBOR) binary 
         and embedded means the original claimset (COSE) with the raw receipt added to the unprotected header
--- a/pyscitt/pyscitt/cli/submit_signed_claims.py
+++ b/pyscitt/pyscitt/cli/submit_signed_claims.py
@@ -73,7 +73,7 @@ def cli(fn):
     )
     parser.add_argument(
         "--receipt-type",
-        choices=[ReceiptType.EMBEDDED.value, ReceiptType.RAW.value],
+        choices=[e.value for e in ReceiptType],
         default=ReceiptType.RAW.value,  # default to raw for backwards compatibility
         help="""
         Downloads the receipt of a given type where raw means a countersignature (CBOR) binary 
--- a/pyscitt/pyscitt/cli/submit_signed_claims.py
+++ b/pyscitt/pyscitt/cli/submit_signed_claims.py
@@ -5,7 +5,7 @@ import argparse
 from pathlib import Path
 from typing import Optional
 
-from ..client import Client
+from ..client import Client, ReceiptType
 from ..verify import StaticTrustStore, verify_receipt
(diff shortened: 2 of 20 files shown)


pyscitt/pyscitt/cli/submit_signed_claims.py
```diff
@@ -48,17 +60,26 @@ def submit_signed_claimset(
...
     group = parser.add_mutually_exclusive_group()
     group.add_argument("--receipt", type=Path, help="Output path to receipt file")
     group.add_argument(
         "--skip-confirmation",
         action="store_true",
         help="Don't wait for confirmation or a receipt",
     )
+    parser.add_argument(
+        "--receipt-type",
+        choices=["embedded", "raw"],
```
andpiccione, line 76: You could use the `ReceiptType` enum values directly here for simplicity:

```suggestion
        choices=[e.value for e in ReceiptType],
```

pyscitt/pyscitt/cli/submit_signed_claims.py
```diff
@@ -48,17 +60,26 @@ def submit_signed_claimset(
...
+        default="raw",  # default to raw for backwards compatibility
```
andpiccione, line 77:
```suggestion
        default=ReceiptType.RAW.value,  # default to raw for backwards compatibility
```

pyscitt/pyscitt/client.py
```diff
@@ -35,6 +35,13 @@ class SigningType(Enum):
     HTTP = "HTTP"
 
 
+class ReceiptType(Enum):
+    """Receipt types supported by the ledger."""
+
+    EMBEDDED = "EMBEDDED"
+    RAW = "RAW"
```
andpiccione, line 42: Assuming you can use the enums directly above, you can convert these to lowercase strings:

```suggestion
    EMBEDDED = "embedded"
    RAW = "raw"
```

pyscitt/pyscitt/cli/submit_signed_claims.py
```diff
@@ -14,27 +14,39 @@ def submit_signed_claimset(
...
     receipt_path: Optional[Path],
+    receipt_type: str,
     service_trust_store_path: Optional[Path],
     skip_confirmation: bool,
 ):
     if path.suffix != ".cose":
-        raise ValueError("unsupported file extension")
+        raise ValueError("unsupported file extension, must end with .cose")
+
+    if receipt_type == "raw":
```
andpiccione, line 24: And then here use the enum values instead of hardcoded strings:

```suggestion
    if receipt_type == ReceiptType.RAW.value:
```

pyscitt/pyscitt/cli/submit_signed_claims.py
```diff
@@ -14,27 +14,39 @@ def submit_signed_claimset(
...
+        r_type = ReceiptType.RAW
+    elif receipt_type == "embedded":
```
andpiccione, line 26: Same here:

```suggestion
    elif receipt_type == ReceiptType.EMBEDDED.value:
```

andpiccione (approved):

demo/github/4-submit.sh
```diff
@@ -5,12 +5,10 @@
...
 
 SCITT_URL=${SCITT_URL:-"https://127.0.0.1:8000"}
-SCITT_TRUST_STORE=tmp/trust_store
 
 TMP_DIR=tmp/github
 
 scitt submit $TMP_DIR/claims.cose \
     --receipt $TMP_DIR/claims.receipt.cbor \
     --url "$SCITT_URL" \
-    --service-trust-store $SCITT_TRUST_STORE \
```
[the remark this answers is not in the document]
ivarprudnikov: Purely because it does not make much sense in verifying in a submit command and the argument is confusing in its name of why it is necessary. There is a separate `validate` command for this.

pyscitt/pyscitt/cli/submit_signed_claims.py
ivarprudnikov: done

pyscitt/pyscitt/verify.py
```diff
@@ -91,10 +91,10 @@ def verify_receipt(
         decoded_receipt = receipt
     else:
         if receipt is None:
-            parsed_receipts = msg.uhdr[COSE_HEADER_PARAM_SCITT_RECEIPTS]
-            # For now, assume there is only one receipt
-            assert len(parsed_receipts) == 1
-            parsed_receipt = parsed_receipts[0]
+            embedded_receipt = crypto.get_last_embedded_receipt_from_cose(buf)
+            if embedded_receipt is None:
```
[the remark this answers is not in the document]
ivarprudnikov: done

pyscitt/pyscitt/cli/submit_signed_claims.py
ivarprudnikov: done

pyscitt/pyscitt/client.py
ivarprudnikov: done

pyscitt/pyscitt/cli/submit_signed_claims.py
ivarprudnikov: done

pyscitt/pyscitt/cli/submit_signed_claims.py
ivarprudnikov: done

demo/github/4-submit.sh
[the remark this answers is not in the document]
andpiccione: Sounds good to me. Then it probably makes sense to directly remove the corresponding CLI argument from the `submit` command so that we avoid users to get confused with its usage.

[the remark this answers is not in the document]
ivarprudnikov: argument removed from submit command as well

PR #215, opened 2024-08-26
Resolve Codeql complaints

ivarprudnikov:
- CodeQL flagged some Python implementation
- This is known to be false positives and I've opened an issue in a respective repo but this did not have any attention so far
- Given we do not want to explain to every person these flags are false positives it is easier to just refactor the code a bit
- RSA key is now hardcoded to be of size 2048, this is used in tests
- EC hash algorithms and keys sizes were also changed to use explicit lookup functions instead of indirect dict that referenced types and values in 3rd party cose library 

To verify if this was resolved I went through hell and fire of reading the CodeQL docs and running it locally:
- Download [CodeQL CLI with all the bells and whistles](https://docs.github.com/en/code-security/codeql-cli/getting-started-with-the-codeql-cli/setting-up-the-codeql-cli)
- Create db from the source files `~/codeql/codeql database create workspace/codeql-dbs --source-root=pyscitt/pyscitt --db-cluster --language=python --overwrite`
- Generate temp token from the GitHub org that has the specific CodeQL queries and set it up in the env vars `export GITHUB_TOKEN=...`
- Run analysis which will fetch the queries using the above token `~/codeql/codeql database analyze workspace/codeql-dbs/python/ microsoft-sdl/python-queries --format=sarif-latest --output=python-results.sarif`
- View the sarif file `python-results.sarif` to see if there are still any issues

### DIFF
--- a/pyscitt/pyscitt/key_vault_sign_client.py
+++ b/pyscitt/pyscitt/key_vault_sign_client.py
@@ -17,12 +17,6 @@ from pyscitt.client import MemberAuthenticationMethod
 
 from . import crypto
 
-ALGORITHMS = {
-    256: ("ES256", "sha256"),
-    384: ("ES384", "sha384"),
-    521: ("ES512", "sha384"),
-}
-
 
 class KeyVaultSignClient(MemberAuthenticationMethod):
     """MemberIdentity implementation that uses Azure Key Vault."""
@@ -120,7 +114,15 @@ class KeyVaultSignClient(MemberAuthenticationMethod):
         pub_key = cert.public_key()
         assert isinstance(pub_key, (EllipticCurvePublicKey))
         key_size = pub_key.curve.key_size
-        signature_algorithm, hash_algorithm = ALGORITHMS[key_size]
+
+        if key_size == 256:
+            signature_algorithm, hash_algorithm = ("ES256", "sha256")
+        elif key_size == 384:
+            signature_algorithm, hash_algorithm = ("ES384", "sha384")
+        elif key_size == 521:
+            signature_algorithm, hash_algorithm = ("ES512", "sha384")
+        else:
+            raise ValueError(f"Unsupported EC size: {key_size}")
 
         digest_to_sign = hashlib.new(hash_algorithm, data).digest()
         sign_result = crypto_client.sign(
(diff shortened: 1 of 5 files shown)


pyscitt/pyscitt/key_vault_sign_client.py
```diff
@@ -120,7 +114,15 @@ def http_sign(self, data: bytes):
...
         assert isinstance(pub_key, (EllipticCurvePublicKey))
         key_size = pub_key.curve.key_size
-        signature_algorithm, hash_algorithm = ALGORITHMS[key_size]
+
+        if key_size == 256:
+            signature_algorithm, hash_algorithm = ("ES256", "sha256")
+        elif key_size == 384:
+            signature_algorithm, hash_algorithm = ("ES384", "sha384")
+        elif key_size == 521:
+            signature_algorithm, hash_algorithm = ("ES512", "sha384")
```
andpiccione, line 123: Shouldn't this be sha512?

```suggestion
            signature_algorithm, hash_algorithm = ("ES512", "sha512")
```

andpiccione (approved):

vimauro (approved):

pyscitt/pyscitt/key_vault_sign_client.py
ivarprudnikov: I tried leaving everything the same as it was without thinking about the values too much but I think your suggestion should be correct.

- [FIPS 186 section 6.1.1](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf) says `It is recommended that the security strength associated with the bit length of n and the security 
strength of the hash function be the same unless an agreement has been made between 
participating entities to use a stronger hash function.` P-521 then should use the strength greater than 256 because bit length of the key is ≥ 512
- [NIST says that P521 is allowed to be used ECDSA see section 3.1](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-186.pdf) and mentions that the hash functions which meet strength ≥ 256 are SHA-512 and SHA3-512 in Table 3.
- [AKV source](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_enums.py#L48C14-L48C19) mentions that P521 is linked to sha512 hash function and references RFC7518, in another place [ES512 implementation is using SHA512](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_internal/algorithms/ecdsa.py#L65)
- [RFC7518](https://datatracker.ietf.org/doc/html/rfc7518) is for JSON web algorithms and also says that `ES512 | ECDSA using P-521 and SHA-512`

PR #219, opened 2024-09-02
WIP fix pretty-receipt command for cose envelopes with embedded receipts

ivarprudnikov:
`pretty-receipt` CLI command fails when presented with the receipt embedded in the cose envelope. Fixing it here to make the command more resilient when parsing arbitrary CBOR. I am fairly sure there is an edge case that is not handled properly at the moment though as JSON encoder does not like pycose objects.

### DIFF
--- a/pyscitt/pyscitt/receipt.py
+++ b/pyscitt/pyscitt/receipt.py
@@ -2,6 +2,7 @@
 # Licensed under the MIT License.
 
 import base64
+import datetime
 import hashlib
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
@@ -23,21 +24,24 @@ HEADER_PARAM_TREE_ALGORITHM = "tree_alg"
 TREE_ALGORITHM_CCF = "CCF"
 
 
+def display_cbor_val(item):
+    print(f"Displaying {type(item)} {item}")
+    if hasattr(item, "__name__"):
+        return item.__name__
+    if type(item) is bytes:
+        return item.hex()
+    if type(item) is datetime:
+        return item.isoformat()
+    return item
+
+
 def hdr_as_dict(phdr: dict) -> dict:
     """
     Return a representation of a list of COSE header parameters that
     is amenable to pretty-printing.
     """
-
-    def display(item):
-        if hasattr(item, "__name__"):
-            return item.__name__
-        if type(item) is bytes:
-            return item.hex()
-        return item
-
     # Decode KID into a 'readable' text string if present.
-    hdr_dict = {display(k): display(v) for k, v in phdr.items()}
+    hdr_dict = {display_cbor_val(k): display_cbor_val(v) for k, v in phdr.items()}
     if hdr_dict.get("KID"):
         hdr_dict["KID"] = bytes.fromhex(hdr_dict["KID"]).decode()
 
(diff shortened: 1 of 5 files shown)


pyscitt/pyscitt/receipt.py
```diff
@@ -21,27 +24,82 @@
...
+    # pycose will use class instances for known and registered headers instead of ints
+    if hasattr(cbor_obj_key, "identifier"):
+        if cbor_obj_key.identifier == crypto.SCITTReceipts.identifier:
+            parsed_receipts = []
+            for item in cbor_obj:
+                if type(item) is bytes:
+                    receipt_as_dict = Receipt.decode(item).as_dict()
+                else:
+                    receipt_as_dict = Receipt.from_cose_obj(item).as_dict()
+            parsed_receipts.append(receipt_as_dict)
```
andpiccione, line 70: Shouldn't this statement be included inside the for loop? I think this is just adding the last converted item to the list.

pyscitt/pyscitt/receipt.py
```diff
@@ -21,27 +24,82 @@
...
-    hdr_dict = {display(k): display(v) for k, v in phdr.items()}
-    if hdr_dict.get("KID"):
-        hdr_dict["KID"] = bytes.fromhex(hdr_dict["KID"]).decode()
+    # attempt to decode nested cbor
+    if type(cbor_obj) is bytes:
+        try:
+            decoded = cbor2.loads(cbor_obj)
+            return cbor_as_dict(decoded, cbor_obj_key)
+        except (CBORError, UnicodeDecodeError):
+            pass
```
andpiccione, line 99: It could make sense to at least print a log of the error/exception here, just for debugging purposes.

pyscitt/pyscitt/receipt.py
```diff
@@ -21,27 +24,82 @@
...
+    elif type(item) is bytes:
+        out = item.hex()
+    return out
+
+
+def cbor_as_dict(cbor_obj: Any, cbor_obj_key: Any = None) -> Any:
     """
-    Return a representation of a list of COSE header parameters that
-    is amenable to pretty-printing.
+    Return a printable representation of a CBOR object.
```
andpiccione, line 58: Is there some risk that some of these conversions may fail with an unhandled exception? If so, it could make sense to add a try-catch to handle possible general exceptions gracefully (either here or upstream when this method is called to print the receipt).

andpiccione (approved):

pyscitt/pyscitt/receipt.py
ivarprudnikov: Thanks! Great observation, the tests could not detect it as we do not have such examples in code, there is not expectation to use signatures with multiple receipts any time soon

pyscitt/pyscitt/receipt.py
ivarprudnikov: Removed this block of coded and will let the bytes to just become a hex, there might be an edge case where it is necessary but not for the current usecases

pyscitt/pyscitt/receipt.py
ivarprudnikov: Added try/catch around receipt parsing, added a default serializer for json serialization to avoid unexpected cbor breaking the command, renamed `cbor_as_dict` to `cbor_to_printable` as it does not only return structs.

Issue #229, opened 2024-10-22
Reproducibility documentation needs updating for confidential containers


achamayou:
The reproducibility doc describes the process for SGX, but not for confidential containers yet,

_Originally posted by @achamayou in https://github.com/microsoft/scitt-ccf-ledger/pull/226#discussion_r1810399793_

ivarprudnikov:
Resolved in https://github.com/microsoft/scitt-ccf-ledger/pull/265

PR #230, opened 2024-10-28
COSE Receipts

achamayou:
Switch to COSE receipts, embedded in unprotected headers of signed statement, as per:

- https://datatracker.ietf.org/doc/draft-ietf-cose-merkle-tree-proofs/
- https://datatracker.ietf.org/doc/draft-birkholz-cose-receipts-ccf-profile/
- https://datatracker.ietf.org/doc/draft-ietf-scitt-architecture/

Also see consolidated (receipt) schema: https://github.com/microsoft/CCF/blob/main/cddl/ccf-receipt.cddl

I have had to disable a few testcases with `mark.skip`, for which I need to investigate possible solutions. All the common paths now work with current-draft-compliant receipts and statements. There is also work to do in future PRs before this can PR to main, including :

- Update Client API/naming to current SCITT
- Update TrustStore to map kid -> key
- Update the CLI


achamayou:
Draft verification logic: https://github.com/microsoft/CCF/pull/6603

app/src/main.cpp
```diff
@@ -57,15 +58,6 @@ namespace scitt
...
 
-  struct DIDFetchContext
-  {
-    std::vector<uint8_t> body;
-    std::string nonce;
-    std::string issuer;
-  };
-  DECLARE_JSON_TYPE(DIDFetchContext);
-  DECLARE_JSON_REQUIRED_FIELDS(DIDFetchContext, body, nonce, issuer);
-
```
achamayou, line 68: Unused after #231

app/src/main.cpp
```diff
@@ -495,6 +436,75 @@ namespace scitt
...
+        // Page 11, vdp is the label for verifiable-proods in the unprotected
+        // header of the receipt
+        int64_t vdp = 396;
+        // -1 is the label for inclusion-proofs
+        auto inclusion_proof = ccf::cose::edit::pos::AtKey{-1};
+        ccf::cose::edit::desc::Value inclusion_desc{
+          inclusion_proof, vdp, *proof};
+
+        auto cose_receipt =
+          ccf::cose::edit::set_unprotected_header(*signature, inclusion_desc);
```
andpiccione, line 481: This part looks the same as in 399-422. I suggest extracting this to a common method to reduce duplication.

app/src/main.cpp
```diff
@@ -495,6 +436,75 @@ namespace scitt
         .set_forwarding_required(ccf::endpoints::ForwardingRequired::Never)
         .install();
 
+      static constexpr auto get_entry_statement_path =
+        "/entries/{txid}/statement";
```
andpiccione, line 440: Don't we want to delete the endpoints returning CBOR receipts (`/entries/{txId}/receipt`) now that we have this new endpoint for returning transparent statements?

test/test_cli.py
```diff
@@ -59,6 +60,7 @@ def test_smoke_test(run, client, tmp_path: Path):
     )
 
 
+@pytest.mark.skip(reason="CLI tests needs fixing")
```
andpiccione, line 63: Let's create an issue or a task to fix these skipped tests in a later PR, just to avoid we forget

app/src/main.cpp
achamayou: It's very inexpensive to expose both statements and receipts, and the latter may be useful if a caller wants to embed multiple receipts in a statement for example, so I'd vote for keeping them both.

test/test_cli.py
achamayou: This is a merge to a branch, not to main, and they're one of the main items before this can PR to main. Happy to have a ticket to track the work though.

test/test_configuration.py
```diff
@@ -300,33 +300,8 @@ def test_cts_hashv_cwtclaims_payload_with_policy(
...
-            cts_hashv_cwtclaims, receipt_type=ReceiptType.EMBEDDED
-        )
+        statement = client.submit_and_confirm(cts_hashv_cwtclaims).receipt_bytes
 
-        # store submission receipt in random file in tmp_path
+        # store statement
         receipt_path = tmp_path / "receipt.cose"
-        receipt_path.write_bytes(submission.receipt_bytes)
-        # print to preview what was accepted and to check if pretty-receipt understands the given receipt
-        main(["pretty-receipt", str(receipt_path)])
```
ivarprudnikov, line 311: `pretty-receipt` must always work, that is why it is included here. Please reinstate

ivarprudnikov (changes requested):

pyscitt/pyscitt/verify.py
```diff
@@ -89,12 +99,33 @@ def verify_receipt(
     decoded_receipt.verify(msg, service_params)
 
 
+def verify_transparent_statement(
```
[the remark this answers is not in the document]
ivarprudnikov: Measurement check was removed, test env should generate a mock measurement which must be present in the receipt headers

pyscitt/pyscitt/receipt.py
```diff
@@ -88,7 +88,7 @@ def cbor_to_printable(cbor_obj: Any, cbor_obj_key: Any = None) -> Any:
         if cbor_obj_key.identifier == X5chain.identifier:
             return [base64.b64encode(cert).decode("ascii") for cert in cbor_obj]
         if cbor_obj_key.identifier == KID.identifier:
-            return cbor_obj.decode()
+            return cbor_obj.hex()
```
ivarprudnikov, line 91: The method is used to pretty-print the cbor structure, we had KID bytes representing a string and decoding seemed to pretty print it well, why change to hex? Did kid change to something which is not a string anymore?

pyscitt/pyscitt/client.py
```diff
@@ -577,6 +577,21 @@ def submit_claim_and_confirm(
         receipt = self.get_receipt(tx, decode=False)
         return Submission(operation_id, tx, receipt, False)
 
+    def submit_and_confirm(
```
ivarprudnikov, line 580: This is a duplicate of `submit_claim_and_confirm` above, why? Surely we can have just one method.

app/src/main.cpp
```diff
@@ -226,51 +254,33 @@ namespace scitt
...
         cose::UnprotectedHeader uhdr;
         try
         {
-          SCITT_DEBUG("Verify submitted claim");
-          std::tie(claim_profile, phdr, uhdr) = verifier->verify_claim(
-            body, ctx.tx, host_time, DID_RESOLUTION_CACHE_EXPIRY, cfg);
+          SCITT_DEBUG("Verify submitted signed statement");
+          std::tie(signed_statement_profile, phdr, uhdr) =
+            verifier->verify_signed_statement(
+              body, ctx.tx, host_time, DID_RESOLUTION_CACHE_EXPIRY, cfg);
```
ivarprudnikov, line 270: if DID:WEB verification is going away why do we still have `DID_RESOLUTION_CACHE_EXPIRY`?

app/src/main.cpp
```diff
@@ -463,27 +438,12 @@ namespace scitt
               historical_state->transaction_id.to_str()));
         }
 
-        SCITT_DEBUG("Get saved SCITT entry");
```
ivarprudnikov, line 466: Leave the debug statement please

app/src/main.cpp
```diff
@@ -495,6 +455,52 @@ namespace scitt
...
+              "Transaction ID {} does not correspond to a submission.",
+              historical_state->transaction_id.to_str()));
+        }
+
+        auto cose_receipt = get_cose_receipt(historical_state->receipt);
+
+        // See https://datatracker.ietf.org/doc/draft-ietf-scitt-architecture/
+        // Page 16, 394 is the label for an array of receipts in the unprotected
+        // header
+        int64_t receipts = 394;
```
ivarprudnikov, line 482: there should be a static variable for this already called `COSE_HEADER_PARAM_SCITT_RECEIPTS`

app/src/main.cpp
```diff
@@ -226,51 +254,33 @@ namespace scitt
...
         }
         catch (const verifier::VerificationError& e)
         {
-          SCITT_DEBUG("Claim verification failed: {}", e.what());
+          SCITT_DEBUG("Signed statement verification failed: {}", e.what());
           throw BadRequestError(errors::InvalidInput, e.what());
         }
-        // Retrieve current enclave measurement of this node
-        // See ccf logic in `/quotes/self`
-        std::string measurement;
```
ivarprudnikov, line 250: Is this implemented in CCF now? how can I reproduce the code that issued the receipt without the measurement?

pyscitt/pyscitt/verify.py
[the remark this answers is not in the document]
achamayou: The code measurement of the TEE that accepted the registration is not included in the SCITT drafts. It's something that can be obtained from the CCF ledger by an auditor if desired.

app/src/main.cpp
achamayou: You can find that information in the ledger, at the transaction where the node was added. Note that the COSE signatures are not signed by node keys, but are signed by the service key instead, which is replicated among nodes.

app/src/main.cpp
achamayou: I don't understand, there's no getting here anymore, why we would log that there is?

pyscitt/pyscitt/client.py
achamayou: Yes, I am in the process of fixing the naming in the test infra and consolidating. It's going to take a while, but do feel free to help.

pyscitt/pyscitt/receipt.py
achamayou: kids for CCF COSE signatures are SHA-256, they are not strings.

test/test_configuration.py
achamayou: Once more, this a PR to a dev branch, some items will happen in later PRs.

pyscitt/pyscitt/verify.py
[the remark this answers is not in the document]
ivarprudnikov: This is part of the feature we added which allows us to show how to reproduce the specific version that issued the receipt. It is not about SCITT drafts because they do not really address reproducibility either. There is no version information in the TLS cert as well, so how does the user know which version of code issued the receipt? The suggestion to look at the ledger files is suboptimal because there is no viable path to it at the moment, AFAIK it exists in theory only.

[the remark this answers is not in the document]
ivarprudnikov: I do not know how to resolve the exact version which issued the receipt, if I miss something then please expand because this is an open issue to make sure SEV-SNP is reproducible and traceable. We need to have concrete steps to link the receipt to the measurement, please update `docs/reproducibility.md` with the steps necessary to link those.

app/src/main.cpp
ivarprudnikov: I've commented above

> You can find that information in the ledger

concrete steps need to be added to the docs/reproducibility.md to avoid confusion and mystery

pyscitt/pyscitt/verify.py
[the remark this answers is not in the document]
achamayou: > The suggestion to look at the ledger files is suboptimal because there is no viable path to it at the moment, AFAIK it exists in theory only.

Do feel free to propose an alternative. I will continue to focus updating the implementation to match the IETF specs.

> We need to have concrete steps to link the receipt to the measurement, please update docs/reproducibility.md with the steps necessary to link those.

This is out of scope for this PR and for this branch.

app/src/main.cpp
achamayou: I do not know which statement you please, please feel free to use the [GitHub suggestion feature](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request) to propose a concrete change.

app/src/main.cpp
achamayou: As explained above, this is out of scope for this change.

pyscitt/pyscitt/verify.py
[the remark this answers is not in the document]
ivarprudnikov: > This is out of scope for this PR and for this branch.

The measurement is being removed in this PR so it should be in scope, I do not understand what is your suggestion, just to forget about it? Is it not possible to update the doc?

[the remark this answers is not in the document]
achamayou: The scope of this PR (which is to a dev branch, not to main) is to switch CTS to COSE receipts. Many things are possible, but are out of scope, including updating the documentation.

To bring some clarity on the topic of having nodes sign their TEE measurement: this never created a security property. The only valuable signature over these measurements is that contained in the hardware attestation.

A relying party can establish all they need to know about the TEE from the attestation, and from the attestation alone, an additional signature by a key generated by the code running inside the TEE creates no additional property. Signing the TEE measurement with a key created by the code is wasteful at best, and dangerous at worst because relying parties may interpret it as an authoritative statement about the code that signed the ledger, which it is not.

Even the node attestation is insufficient to replay/audit a single signature transaction, because the code is only one of the inputs, the state of the ledger/KV is the other. The state of the ledger (captured in the Merkle Proof) contains all past node attestations, and is always sufficient for audit purposes. Extracting an unauthenticated measurement from it does not create a additional security property.

Finally, a node attestation has limited usefulness for audit, even with node-scoped keys, without establishing the history of the service key used to endorse it. This is particularly true for a pure-integrity application like CTS, which aims to produce a single, unequivocal history. Adding an unauthenticated measurement to signatures neither improves the full audit story, nor partial consistency checks based on a set of receipts.

app/src/main.cpp
```diff
@@ -463,27 +437,12 @@ namespace scitt
               historical_state->transaction_id.to_str()));
         }
 
-        SCITT_DEBUG("Get saved SCITT entry");
-        auto entry_info_table =
-          historical_tx.template ro<EntryInfoTable>(ENTRY_INFO_TABLE);
-        auto entry_info = entry_info_table->get().value();
+        auto cose_receipt = get_cose_receipt(historical_state->receipt);
```
ivarprudnikov, line 440:
```suggestion
        SCITT_DEBUG("Get signed statement from the ledger");
        auto cose_receipt = get_cose_receipt(historical_state->receipt);
```

app/src/main.cpp
achamayou: https://github.com/microsoft/scitt-ccf-ledger/issues/229#issue-2605382283

PR #231, opened 2024-11-14
Removed did:web support

vimauro:
Removed did:web support

TODO:
1 test_submit_claim is still failing


app/src/main.cpp
```diff
@@ -429,15 +361,7 @@ namespace scitt
...
-        try
-        {
-          post_entry_common(ctx, host_time, body);
-        }
-        catch (const did::web::AsyncResolutionNeeded& e)
-        {
-          start_asynchronous_registration(ctx, host_time, e.did, body);
-          return;
-        }
+        post_entry_common(ctx, host_time, body);
```
andpiccione, line 364: nit: I believe we could also remove the `post_entry_continuation`, since it was mainly used to continue the `post_entry` logic after an asynchronous `did:web` resolution.

Similarly, the `post_entry_common` can probably be merged inside this controller since it won't be called anywhere else. But we can still keep that in a separate method if we prefer that.

app/src/main.cpp
```diff
@@ -429,15 +361,7 @@ namespace scitt
             "Failed to get host time: {}", ccf::api_result_to_str(result)));
         }
 
-        try
-        {
-          post_entry_common(ctx, host_time, body);
-        }
-        catch (const did::web::AsyncResolutionNeeded& e)
-        {
-          start_asynchronous_registration(ctx, host_time, e.did, body);
```
andpiccione, line 438: I think we should also be able to remove all the [operations endpoints](https://github.com/microsoft/scitt-ccf-ledger/blob/4f4eb5310afb52c17e7fd4832c3d7d3bdbfd2e1d/app/src/operations_endpoints.h#L525), since they were primarily used to check the status of a post entry operation when there was a did:web resolution.

build.sh
```diff
@@ -35,17 +35,6 @@ install_dir=/tmp/scitt
...
-    -DCMAKE_BUILD_TYPE="${CMAKE_BUILD_TYPE}" \
-    -DCMAKE_INSTALL_PREFIX=$install_dir \
-    -DCOMPILE_TARGET="${ATTESTED_FETCH_PLATFORM}" \
-    "$root_dir/3rdparty/attested-fetch"
-
-ninja -C build/attested-fetch ${NINJA_FLAGS} --verbose
-ninja -C build/attested-fetch ${NINJA_FLAGS} install
-
 if [ "$PLATFORM" = "sgx" ]; then
     ATTESTED_FETCH_MRENCLAVE_HEX=$(/opt/openenclave/bin/oesign dump -e $install_dir/libafetch.enclave.so.signed | sed -n "s/mrenclave=//p")
```
ivarprudnikov, line 39: This should go as well as the idea was to register the mrenclave from the attested fetch binary in the ledger application as a trusted enclave that it can communicate with. So the code paths that rely on `ATTESTED_FETCH_MRENCLAVE_HEX` need to be removed as well

app/src/did/attested.h
ivarprudnikov, line 1: I think this needs to go as well as attested fetch is explicitly used for did:web resolution

ivarprudnikov:
`libcurl4-openssl-dev` dependency should also be removed as it is something that `attested-fetch` depends on

app/src/main.cpp
achamayou: +1 for the merge

achamayou (approved):

app/src/main.cpp
vimauro: Agreed, I've merged them and removed `post_entry_continuation`

app/src/main.cpp
```diff
@@ -267,17 +259,7 @@ namespace scitt
...
-          SCITT_DEBUG("No policy applied, but CWT issuer present");
-          throw BadRequestError(
-            errors::PolicyFailed,
-            "Policy was not met: CWT issuer present but no policy configured");
-        }
-        else
-        {
-          SCITT_DEBUG("No policy applied");
-        }
+        SCITT_DEBUG("No policy applied");
```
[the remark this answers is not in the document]
vimauro: Restored the initial behavior, it was removed by mistake during the cleanup

app/src/main.cpp
vimauro: I suppose I still need to port some logic from the `operations_endpoints` into the post_entry endpoint since the response will be empty otherwise, right ?
e.g. `operation_locally_committed_func`

ivarprudnikov: Operations endpoints are standard handlers in azure to deal with async ops, sdk depends on them at the moment, if you were to remove those then the response needs to match the expectations in SCRAPI SCITT API proposals I think, @achamayou could suggest here
scrapi spec draft https://github.com/ietf-wg-scitt/draft-ietf-scitt-scrapi/blob/main/draft-ietf-scitt-scrapi.md#register-signed-statement

ivarprudnikov: From Azure API perspective post must return the location of the created resource as a URL https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#http-return-codes just like in the case of SCRAPI, it is just SCRAPI suggests the response be in `Payload (in CBOR diagnostic notation)` which is kind of weird given they also use JSON for errors.

achamayou: Interestingly, Darrel Miller has written a review of the SCITT architecture containing detailed views on errors and long running operations: https://datatracker.ietf.org/doc/review-ietf-scitt-architecture-01-httpdir-early-miller-2023-04-24/

For now I think we:

- Keep the handling for async commit as is
- Remove the did resolution handling

Once we've moved over to the new receipt format, we can do a pass over the exact details of the API and errors.

vimauro: I've restored the operation endpoints, kept the same data structures and removed DID:Web async resolution

achamayou (approved):

PR #246, opened 2024-12-03
remove remains of sgx and update reproducibility doc

ivarprudnikov:
- removes remains of sgx usage in code
- removes openenclave
- updates reproducibility doc and adds TODOs as the reproducibility story on SNP is not finished yet
- removes verification of deprecated actions from constitution
- bumps up pyscitt version


app/constitution/actions.js
```diff
@@ -797,22 +797,6 @@ const actions = new Map([
         checkType(args.issuer, "string", "issuer");
         checkType(args.auto_refresh, "boolean?", "auto_refresh");
         checkType(args.ca_cert_bundle_name, "string?", "ca_cert_bundle_name");
-        checkEnum(args.key_filter, ["all", "sgx"], "key_filter");
```
andpiccione, line 800: You should also remove other SGX-related actions from the constitution, like `add_node_code`, `remove_node_code`, etc.

docker/README.md
```diff
@@ -17,12 +17,3 @@ PLATFORM="virtual" ./docker/run-dev.sh
...
 Both scripts accept different variables for customization. Please refer to the corresponding scripts for the full list of available variables to use.
-
-## Build and push docker images to a container registry
-
-After building a docker image, you can push it to a container registry (e.g., Azure Container Registry). You can use the `push_image.sh` script to automatically build and push the image to ACR:
-
-```bash
-# Build docker image for SGX
-./docker/push_image.sh
-```
```
andpiccione, line 28: Why removing this entire section?

docker/push_image.sh
```diff
@@ -1,38 +0,0 @@
-#!/bin/bash
```
andpiccione, line 1: I don't understand the need to remove this entire file, it is useful for manually pushing an image to ACR

docs/reproducibility.md
```diff
@@ -10,53 +10,42 @@ You need a couple pieces of information to begin with:
...
-    ```
+The details of how to reproduce the Guest VM are not ready yet. The source code will be open sourced at some point and it will be possible to build from it to reproduce the `measurement` in the report. It will make sure that the security policy enforcement logic is as expected.
+
+### Security policy and container image
 
-- As you can see in the example, the value printed in the build output matched the one from a running application.
\ No newline at end of file
+`Host data` contains the hash of the security policy, so we should be able to:
+- (TODO) build the docker image and create the same policy from it, or
+- (TODO) download the policy using the HTTP API, once it is available
```
andpiccione, line 51: The plain text policy is already available in the CCF SNP host data (it is an optional value and can be left empty, but we will always add it during code upgrades).

docs/reproducibility.md
```diff
@@ -10,53 +10,42 @@ You need a couple pieces of information to begin with:
...
-    96c40e2532ba329849e7fede3f3d888a0423a1dc0f3d0511b138617cc3aa9e94
-    ```
+The details of how to reproduce the Guest VM are not ready yet. The source code will be open sourced at some point and it will be possible to build from it to reproduce the `measurement` in the report. It will make sure that the security policy enforcement logic is as expected.
+
+### Security policy and container image
 
-- As you can see in the example, the value printed in the build output matched the one from a running application.
\ No newline at end of file
+`Host data` contains the hash of the security policy, so we should be able to:
+- (TODO) build the docker image and create the same policy from it, or
```
andpiccione, line 50: This is not possible without the pod yaml or exact deployment configurations, which are not open source.

docs/reproducibility.md
andpiccione, line 51: I would honestly just remove this part as it contains confusing information. We can populate it once we agree on the exact policy reproducibility steps on SNP.

pyscitt/setup.py
```diff
@@ -6,7 +6,7 @@
 from setuptools import find_packages, setup
 
 PACKAGE_NAME = "pyscitt"
-PACKAGE_VERSION = "0.5.0"
+PACKAGE_VERSION = "1.0.0"
```
andpiccione, line 9: Any reason for making this the first stable release?

docs/reproducibility.md
```diff
@@ -10,53 +10,42 @@ You need a couple pieces of information to begin with:
...
     $ curl -s --cacert cacert.pem https://<LEDGER-URL>/app/version | jq ".scitt_version"
-    "0.7.2"
+    "1.2.3"
     ```
 
-## Reproduce measurement
+### Extract measurements
 
-To reproduce the same measurement do a docker build locally using the expected build image from [`microsoft/CCF`](https://github.com/microsoft/ccf):
+The quote contains the attestation report that has the necessary measurements. `Measurement` will have the launch measurement of the guest virtual machine (aka utility VM or UVM) which was used to run the container. `Host data` will have the security policy measurement which was used to make sure no other container was launched except the one you are inspecting, it will contain container image layer hashes.
```
achamayou, line 27: Security policies do not necessarily contain image layer digests, they may contain fragment descriptors instead, so we probably want to be a little more generic here.

docs/reproducibility.md
```diff
@@ -10,53 +10,42 @@ You need a couple pieces of information to begin with:
...
-        FROM ghcr.io/microsoft/ccf/app/dev/sgx:ccf-"$CCF_VERSION" git config --global --add safe.directory "*" && ./docker/build.sh
-    ```
-- The build will print the value of `MRENCLAVE` in the output, similar to:
+### Guest VM
 
-    ```sh
-    mrenclave.txt
-    96c40e2532ba329849e7fede3f3d888a0423a1dc0f3d0511b138617cc3aa9e94
-    ```
+The details of how to reproduce the Guest VM are not ready yet. The source code will be open sourced at some point and it will be possible to build from it to reproduce the `measurement` in the report. It will make sure that the security policy enforcement logic is as expected.
```
achamayou, line 45: I recommend against making forward-looking statements in the documentation, even when they seem reasonable (as is the case here), describing exactly what exists today is best, and it can be updated when changes take place.

achamayou (approved):

pyscitt/setup.py
achamayou: I think holding on until we get a little more user feedback would be best. We could move to 0.6.0?

docker/README.md
ivarprudnikov: This whole page needs to be updated and the building of an image alone does not help. And, I am not even sure if this is the right way to build anymore so it needs to be rewritten anyway.

docker/push_image.sh
ivarprudnikov: it is not used

docs/reproducibility.md
ivarprudnikov: I do not understand what do you mean, are you saying anyone can download it to compare the policy hash that exist in the report to the hash of the plain text policy?

pyscitt/setup.py
ivarprudnikov: CLI contains breaking changes, please suggest something else

docker/README.md
andpiccione: This section is just for pushing an image to a container registry and it can be useful for local development and testing. Unless there's a specific reason to remove it (I don't see anything SGX-related in here, except for the comment line), I would not remove it.

docs/reproducibility.md
ivarprudnikov: Please suggest the text change

docker/push_image.sh
andpiccione: As said, it can be useful for local development and for quickly pushing an image to a container registry without having to use the release pipeline. Unless there's a specific reason to remove it, I would not delete it.

docs/reproducibility.md
ivarprudnikov: Please suggest an alternative as otherwise we'll just go back and forth

docs/reproducibility.md
ivarprudnikov: removed the suggestions

app/constitution/actions.js
ivarprudnikov: removed `add_node_code`, `remove_node_code` please comment if anything else needs to be removed

docs/reproducibility.md
andpiccione: The policy hash and the plain text policy accepted by a CCF instance can be retrieved from the CCF endpoint `/gov/kv/nodes/snp/host_data`, which contains a map {policy_hash -> policy_text} of security policies that are allowed by the service. The policy hash can be compared to the one in the SNP report produced locally, while from the plain text policy it is possible to extract the container image layers that can be verified to match a locally-built image.

docker/README.md
ivarprudnikov: Oh wait sorry I thought you were referring to another page. This piece of code and docs is not used anymore, anyone can push the docker image to their registry and there is no need to have a script to do that. The less we have to maintain the better

pyscitt/setup.py
andpiccione: Until a first stable release, it is possible to increment the minor version number even for breaking changes. This is standard for semantic version numbers used for software versioning. See https://semver.org/#spec-item-4

andpiccione: (FWIW, we already had breaking changes in the CLI in the past and we always incremented the minor number accordingly)

docs/reproducibility.md
```diff
@@ -10,53 +10,40 @@ You need a couple pieces of information to begin with:
...
```
achamayou, line 45:
```suggestion
Guest VM measurements can be authenticated using [platform endorsements](https://github.com/microsoft/confidential-aci-examples/blob/main/docs/Confidential_ACI_SCHEME.md#reference-info-base64).
```

docs/reproducibility.md
ivarprudnikov: simplified

docs/reproducibility.md
ivarprudnikov: simplified

docs/reproducibility.md
```diff
@@ -10,53 +10,40 @@ You need a couple pieces of information to begin with:
...
     $ curl -s --cacert cacert.pem https://<LEDGER-URL>/app/version | jq ".scitt_version"
-    "0.7.2"
+    "1.2.3"
     ```
 
-## Reproduce measurement
+### Extract measurements
 
-To reproduce the same measurement do a docker build locally using the expected build image from [`microsoft/CCF`](https://github.com/microsoft/ccf):
+The quote contains the attestation report that has the necessary measurements. `Measurement` will have the launch measurement of the guest virtual machine (aka utility VM or UVM) which was used to run the container. `Host data` will have the security policy measurement which was used to make sure no other container was launched except the one you are inspecting, it will contain container image layer hashes.
```
achamayou:
```suggestion
The quote contains the attestation report that has the necessary measurements. `Measurement` will have the launch measurement of the guest virtual machine (aka utility VM or UVM) which was used to run the container. `Host data` will have the security policy measurement which was used to make sure no other container was launched except the one you are inspecting. More detail on implementing relying party logic can be [found here](https://github.com/microsoft/confidential-aci-examples/blob/main/docs/Confidential_ACI_SCHEME.md#reference-info-base64).
```

docs/reproducibility.md
ivarprudnikov: I'll work around it as users do not have access to `reference-info-base64` but `uvm_endorsements` in the quote

docs/reproducibility.md
ivarprudnikov: updated

pyscitt/setup.py
ivarprudnikov: done

PR #252, opened 2024-12-12
Document standard alignment of the implementation

achamayou:
Document what is aligned, and what remains to be aligned.

For remaining elements, let's keep track of intent in this document.


docs/scitt.md
```diff
@@ -0,0 +1,126 @@
...
+# Register Signed Statement
+
+See Section 2.1.2
+
+Request
+
+```
+POST /entries HTTP/1.1
+Host: transparency.example
+Accept: application/json
```
ivarprudnikov, line 66: Accept should be `application/cose` here, or omitted

docs/scitt.md
```diff
@@ -0,0 +1,126 @@
...
+
+Content-Type: application/cbor
+Retry-After: <seconds>
+
+{
+
+  "identifier": TBD,
+
+}
+```
```
ivarprudnikov, line 93: This should be removed as we do not do async anymore, and even if we do repeating a post with `Retry-After` does not make sense

ivarprudnikov: I think we should just ignore that `Retry-After` header as it makes no sense after post request

docs/scitt.md
achamayou, line 66:
```suggestion
Accept: application/cose
```

docs/scitt.md
achamayou: I am not sure I understand, do you mean CTS should not set a Retry-After on 202 responses?

ivarprudnikov: Correct, as we do not want clients retrying the post request.

docs/scitt.md
```diff
@@ -0,0 +1,125 @@
...
+
+scitt-ccf-ledger returns Transparent Statement outputs as specified in Section 4.4 of the [Architecture Draft](https://datatracker.ietf.org/doc/draft-ietf-scitt-architecture/). See tests for details.
+
+# Receipts
+
+scitt-ccf-ledger returns COSE Receipts, either standalone or embedded in Transparent Statement outputs, as specified in Section 4 of the [COSE Receipt Draft](https://datatracker.ietf.org/doc/draft-ietf-cose-merkle-tree-proofs/), using proofs that follow the [CCF Tree algorithm profile Draft](https://datatracker.ietf.org/doc/draft-birkholz-cose-receipts-ccf-profile/).
+
+# API
+
+scitt-ccf-ledger does not currently implement the [SCITT Reference API](https://github.com/ietf-wg-scitt/draft-ietf-scitt-scrapi/blob/main/draft-ietf-scitt-scrapi.md). The following changes are necessary to align:
```
ivarprudnikov, line 27: FYI receipts endpoint is missing in the list

achamayou: There are only two mandatory endpoints in the current draft, the transparency configuration, and registration which can be async and return the location at which the receipt is eventually available. I think that's the scheme we are going for at the moment, but you're right that we can specify that too (I think it's a `GET application/cose`).

ivarprudnikov: Fair enough, let's not do that now as it is not mandatory

docs/scitt.md
```diff
@@ -0,0 +1,125 @@
...
+
+```
+HTTP/1.1 200 Ok
+Content-Type: application/cbor
+
+Payload (in CBOR diagnostic notation)
+
+{
+    "issuer": "https://transparency.example",
+    "jwks_uri": "https://transparency.example/jwks",
```
achamayou, line 51:
```suggestion
    "jwksUri": "https://transparency.example/jwks",
```

ivarprudnikov:
Not sure if it is worth keeping after the changes were applied in #272

achamayou:
@ivarprudnikov I think it's worth having, because the question comes up sometimes, and it's good to have a centralised answer. I have updated it the document, following changes in #272 and using specific Draft versions.

andpiccione (approved):

Issue #187, opened 2024-03-19
Inconsistent use of JSON keys snake_case and camelCase in responses


ivarprudnikov:
Inconsistent use of JSON keys: `snake_case` vs `camelCase`

Configuration object is using `snake_case` and is exposed to the user. The `/entries/txIds` endpoint returns a list that has `nextLink`, see renaming logic in JSON https://github.com/microsoft/scitt-ccf-ledger/blob/5cea879c7256a1ce174e3c21d9a25f580ec543aa/app/src/call_types.h#L75-L78 

**Expected behavior**
All JSON keys should use `camelCase` naming convention.

achamayou:
This is consistent with the current state of the spec: https://github.com/ietf-wg-scitt/draft-ietf-scitt-scrapi/blob/main/draft-ietf-scitt-scrapi.md#transparency-configuration, although the layout of the config object has changed and moved to CBOR since.

ivarprudnikov:
Fixed that in the Azure REST API spec but need changes in the code https://github.com/Azure/azure-rest-api-specs/pull/31896

ivarprudnikov:
fixed with in #256

PR #256, opened 2024-12-19
Use consistent camelCase in http api

ivarprudnikov:
Issue #187 

- rename json fields as per https://github.com/Azure/azure-rest-api-specs/pull/31896 
- remove dead code related to unused did:web resolver and signature issuance


app/src/kv_types.h
```diff
@@ -156,33 +144,44 @@ namespace scitt
     Policy policy = {};
     Authentication authentication = {};
 
-    // The long-term stable identifier of this service, as a DID.
+    // The long-term stable issuer string of this service.
     // If set, it will be used to populate the issuer field of receipts
-    std::optional<std::string> service_identifier;
+    std::optional<std::string> service_issuer;
```
achamayou, line 149: Same remark as above, `cose_signatures.issuer` is used for that purpose, not this value.

test/infra/fixtures.py
```diff
@@ -306,16 +305,17 @@ def pytest_collection_modifyitems(config, items):
...
+    Get the long term service issuer string.
 
     The service is configured to include this identifier in receipts.
     """
-
     result = urlparse(service_url)
     assert result.hostname is not None
-    return format_did_web(result.hostname, result.port)
+    if result.port is None or result.port == 443 or result.port == 80:
+        return "https://" + result.hostname
```
achamayou, line 317: if result.port == 80 it's arguable that this should return "http://" + result.hostname instead.

app/constitution/scitt.js
```diff
@@ -12,31 +12,25 @@ actions.set("set_scitt_configuration",
...
+        checkType(args.configuration.authentication.allowUnauthenticated, "boolean?", "configuration.authentication.allowUnauthenticated");
         checkType(args.configuration.authentication.jwt, "object?", "configuration.authentication.jwt");
         if (args.configuration.authentication.jwt) {
-          checkType(args.configuration.authentication.jwt.required_claims, "object?", "configuration.authentication.jwt.required_claims");
+          checkType(args.configuration.authentication.jwt.requiredClaims, "object?", "configuration.authentication.jwt.requiredClaims");
         }
       }
 
-      checkType(args.configuration.service_identifier, "string?", "configuration.service_identifier");
+      checkType(args.configuration.serviceIssuer, "string?", "configuration.serviceIssuer");
```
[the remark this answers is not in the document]
ivarprudnikov: left it but marked the property in KV as deprecated in the comment. It is still used in the did:web part of the code that generated the page but should be removed once we merge the transparency-config + jwks

app/src/kv_types.h
ivarprudnikov: add a section in the doc about `cose_signatures` and removed  `service_issuer` from python tests and from docs

test/infra/fixtures.py
ivarprudnikov: this block was removed

PR #261, opened 2025-01-30
Remove support for Notary signatures

achamayou:
The service is focusing on handling SCITT signatures consistent with https://datatracker.ietf.org/doc/draft-ietf-scitt-architecture/ and https://datatracker.ietf.org/doc/draft-ietf-scitt-scrapi/.


ivarprudnikov (approved):
needs to be synced with master due to ccf version changes in `main`

achamayou:
> needs to be synced with master due to ccf version changes in `main`

done

PR #265, opened 2025-02-11
Doc/reproducibility in amd sev snp

ivarprudnikov:
- update reproducibility doc to explain how to compare image layers that are present in the security layer
- remove unused Python installation from docker images
- switch to using more detailed version via `git describe --tags --long`, e.g. `0.12.1-2-g0b45e35`

### DIFF
--- a/docs/reproducibility.md
+++ b/docs/reproducibility.md
@@ -8,48 +8,136 @@ The assumption here is that the original build was done using a Docker.
 
 You need a couple pieces of information to begin with:
 
-- The ledger certificate. It might be distributed in a variety of ways by the ledger operator, please follow their guidance. Otherwise it is accessible at `https://<LEDGER-URL>/app/parameters`.
-
-- The quote of a running application code, get it from `https://<LEDGER-URL>/node/quotes/self` (replace <LEDGER-URL> with the URL of your ledger), e.g.:
+- The ledger certificate. It might be distributed in a variety of ways by the ledger operator, please follow their guidance. Otherwise it is accessible at `https://<LEDGER-URL>/app/parameters`, e.g.:
 
     ```sh
-    $ curl -s --cacert cacert.pem https://<LEDGER-URL>/node/quotes/self > node-quote.json
+    $ curl -k "https://<LEDGER-URL>/app/parameters" | jq -r .serviceCertificate | base64 -d > cacert.der
+    $ openssl x509 -inform der -in cacert.der > cacert.pem
     ```
+
+- The quote of a running application code, get it from `https://<LEDGER-URL>/node/quotes` which will contain the measurements of each node in the network. They will be the same almost all of the time except when upgrading to the new version e.g.:
+
+    ```sh
+    $ curl -s --cacert cacert.pem https://<LEDGER-URL>/node/quotes > node-quotes.json
+    ```
+
 - Source code version which was used to build the application, usually found in `https://<LEDGER-URL>/app/version`. If `app/version` is ambiguous then check the build logs of the Docker image to understand which commit produced the candidate value. e.g., `fb2c496416fbab20837fedda0ba6db58d819fa5f5c1b3916062eb2fb9d889966` was built from `fd77c0c69ee890bdc2fcf6ef0c9dddb7b211e164`.
 
     ```sh
     $ curl -s --cacert cacert.pem https://<LEDGER-URL>/app/version | jq ".version"
-    "1.2.3"
+    "0.11.0"
     ```
 
-### Extract measurements
+- Security policy used to verify the container image ([ccf docs](https://microsoft.github.io/CCF/main/governance/gov_api_schemas/2024-07-01.html#get--gov-service-join-policy)), it will contain image layers, e.g.:
+
+    ```sh
+    $ curl -s --cacert cacert.pem https://<LEDGER-URL>/gov/service/join-policy?api-version=2024-07-01 > service-join-policy.json
+    ```
+
+### Extract measurements from the report
 
 The quote contains the attestation report that has the necessary measurements. `Measurement` will have the launch measurement of the guest virtual machine (aka utility VM or UVM) which was used to run the container. `Host data` will have the security policy measurement which was used to make sure no other container was launched except the one you are inspecting. More detail on implementing relying party logic can be [found in Confidential ACI scheme documentation](https://github.com/microsoft/confidential-aci-examples/blob/main/docs/Confidential_ACI_SCHEME.md#reference-info-base64).
 
 - Decode the report and save it to a file:
 
     ```sh
-    cat node-quote.json | jq -r '.raw' | base64 -d > snp-report.bin
+    $ cat node-quotes.json | jq -r '.quotes[0].raw' | base64 -d > snp-report.bin
     ```
 
-- Inspect the report to get the measurements. You could use https://github.com/virtee/snpguest to display the report details:
+- Inspect the report to get the measurements. You could use https://github.com/virtee/snpguest to display the report details, e.g.:
 
     ```sh
-    snpguest display report snp-report.bin
+    $ curl -LO https://github.com/virtee/snpguest/archive/refs/tags/v0.8.0.tar.gz
+    $ tar -xvf v0.8.0.tar.gz
+    $ cd snpguest-0.8.0
+    $ cargo build -r
+    $ cd ..
+    $ ./snpguest-0.8.0/target/release/snpguest display report snp-report.bin
+    <...>
+    Measurement:
+    18 25 a4 bf 2a 9c 38 35 66 a7 17 63 26 83 9a c0
+    e3 6a 1c 5b 37 e9 e6 fa bc 8f dd 71 30 d5 8c ef
+    56 f4 34 75 02 b9 47 89 53 0c ec 19 8a a5 15 43
+
+    Host Data:
+    5a e7 b1 4e e0 c9 c4 fe 26 7d 19 1f 25 b2 0f ff
+    e2 4e 29 c4 ac 41 9c 50 50 1d 20 c8 69 bb ba 65
+    <...>
     ```
+
+- You could also verify the provided report with services such as Microsoft Azure Attestation Service, this step is excluded for the brevity reasons
+
+### Extract image layers from security policy
+
+Inspect the service join policy content anextract the Rego policy used to validate the container:
+
+```sh
+$ cat service-join-policy.json | jq -r '.snp.hostData["5ae7b14ee0c9c4fe267d191f25b20fffe24e29c4ac419c50501d20c869bbba65"]' | printf "%s" "$(cat)" > ccepolicy.rego
+```
+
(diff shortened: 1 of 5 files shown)


docs/reproducibility.md
```diff
@@ -8,48 +8,152 @@ The assumption here is that the original build was done using a Docker.
...
+
+You can use an OPA agent to convert the Rego file to JSON to then select layers or do that manually, e.g.:
+
+```sh
+$ curl -s -LO https://openpolicyagent.org/downloads/v1.1.0/opa_linux_amd64_static
+$ chmod 755 opa_linux_amd64_static
+$ ./opa_linux_amd64_static parse ccepolicy.rego -f json > ccepolicy.json
+
+# jq find the containers rule and extract layers of each container
+$ cat ccepolicy.json | jq '[ .rules[] | select(.head.name == "containers") | .head.value.value[].value[] | select(.[].value == "layers") | .[1].value | map(.value) ]' > containerlayers.json
```
andpiccione, line 90: I believe the field where the image layers are specified in not the same across different SNP platforms (C-ACI and C-AKS). I suggest specifying this only applies to C-ACI security policies generated using the `az confcom acipolicygen` CLI.

docs/reproducibility.md
```diff
@@ -8,48 +8,152 @@ The assumption here is that the original build was done using a Docker.
...
     ```
 
+- You could also verify the provided report with services such as Microsoft Azure Attestation Service, this step is excluded for the brevity reasons
+
+### Extract image layers from security policy
+
+Inspect the service join policy content and extract the Rego policy used to validate the container:
+
+```sh
+$ cat service-join-policy.json | jq -r '.snp.hostData["5ae7b14ee0c9c4fe267d191f25b20fffe24e29c4ac419c50501d20c869bbba65"]' | printf "%s" "$(cat)" > ccepolicy.rego
```
andpiccione, line 75: There could be multiple accepted security policies in the join policy (especially during code upgrades). How do customers know which one they should verify? I guess they might need to repeat the same process for every policy in the list until they find one matching the host data present in the attestation report.

docs/reproducibility.md
```diff
@@ -8,48 +8,152 @@ The assumption here is that the original build was done using a Docker.
...
-- The ledger certificate. It might be distributed in a variety of ways by the ledger operator, please follow their guidance. Otherwise it is accessible at `https://<LEDGER-URL>/app/parameters`.
+- The ledger certificate. It might be distributed in a variety of ways by the ledger operator, please follow their guidance. Otherwise it is accessible at `https://<LEDGER-URL>/app/parameters`, e.g.:
 
-- The quote of a running application code, get it from `https://<LEDGER-URL>/node/quotes/self` (replace <LEDGER-URL> with the URL of your ledger), e.g.:
+    ```sh
+    $ curl -k "https://<LEDGER-URL>/app/parameters" | jq -r .serviceCertificate | base64 -d > cacert.der
+    $ openssl x509 -inform der -in cacert.der > cacert.pem
+    ```
+
+- The quote of a running application code, get it from `https://<LEDGER-URL>/node/quotes` which will contain the measurements of each node in the network. They will be the same almost all of the time except when upgrading to the new version e.g.:
```
achamayou, line 18: Filed https://github.com/microsoft/CCF/issues/6822, I will make the update once this is available in a release.

docs/reproducibility.md
```diff
@@ -8,48 +8,152 @@ The assumption here is that the original build was done using a Docker.
...
 - Source code version which was used to build the application, usually found in `https://<LEDGER-URL>/app/version`. If `app/version` is ambiguous then check the build logs of the Docker image to understand which commit produced the candidate value. e.g., `fb2c496416fbab20837fedda0ba6db58d819fa5f5c1b3916062eb2fb9d889966` was built from `fd77c0c69ee890bdc2fcf6ef0c9dddb7b211e164`.
 
     ```sh
     $ curl -s --cacert cacert.pem https://<LEDGER-URL>/app/version | jq ".version"
-    "1.2.3"
+    "0.12.3-0-gaaaaaaa"
     ```
 
-### Extract measurements
+- Security policy used to verify the container image ([ccf docs](https://microsoft.github.io/CCF/main/governance/gov_api_schemas/2024-07-01.html#get--gov-service-join-policy)), it will contain image layers, e.g.:
```
achamayou, line 31:
```suggestion
- Join policy used to verify the container image ([ccf docs](https://microsoft.github.io/CCF/main/governance/gov_api_schemas/2024-07-01.html#get--gov-service-join-policy)), it will contain image layers, e.g.:
```

The join policy is what the primary applies to a joiner's attestation to decide if it can join or not. It does not directly contain container layer digests, rather it contains a set of digests of confidential container security policies that are acceptable (`snp.hostData`). Details can be found here: https://microsoft.github.io/CCF/main/governance/gov_api_schemas/2024-07-01.html#get--gov-service-join-policy

The confidential container security policies themselves do contain the digests of layers.

docs/reproducibility.md
```diff
@@ -8,48 +8,152 @@ The assumption here is that the original build was done using a Docker.
...
 
 - Decode the report and save it to a file:
 
     ```sh
-    cat node-quote.json | jq -r '.raw' | base64 -d > snp-report.bin
+    $ cat node-quotes.json | jq -r '.quotes[0].raw' | base64 -d > snp-report.bin
     ```
 
-- Inspect the report to get the measurements. You could use https://github.com/virtee/snpguest to display the report details:
+- Inspect the report to get the measurements. You could use https://github.com/virtee/snpguest to display the report details, e.g.:
```
achamayou, line 47: There's also a hex2report tool in https://github.com/microsoft/confidential-sidecar-containers/blob/main/tools/get-snp-report/hex2report.c, mentioned in passing in https://github.com/microsoft/confidential-sidecar-containers?tab=readme-ov-file#fetching-an-attestion-report.

docs/reproducibility.md
```diff
@@ -8,48 +8,152 @@ The assumption here is that the original build was done using a Docker.
...
+    e3 6a 1c 5b 37 e9 e6 fa bc 8f dd 71 30 d5 8c ef
+    56 f4 34 75 02 b9 47 89 53 0c ec 19 8a a5 15 43
+
+    Host Data:
+    5a e7 b1 4e e0 c9 c4 fe 26 7d 19 1f 25 b2 0f ff
+    e2 4e 29 c4 ac 41 9c 50 50 1d 20 c8 69 bb ba 65
+    <...>
     ```
 
+- You could also verify the provided report with services such as Microsoft Azure Attestation Service, this step is excluded for the brevity reasons
```
achamayou, line 68:
```suggestion
- You could also verify the provided report with services such as Microsoft Azure Attestation Service, this step is excluded for brevity reasons
```

docs/reproducibility.md
achamayou: If the intention is to audit the running service, they all need to be checked. If the intention is to verify a single node, in principle they could start from `/quote/self` and check that, but the load balancer will make that somewhat meaningless.

We could add a `/quote/{node_id}` to facilitate auditing a specific node, but I think in practice it is simpler and more thorough to recommend auditing all the nodes.

docs/reproducibility.md
```diff
@@ -8,48 +8,152 @@ The assumption here is that the original build was done using a Docker.
...
+
+**Note:** image layers in the security policy use [dmverity hashes](https://www.kernel.org/doc/html/latest/admin-guide/device-mapper/verity.html), hence you will need to convert the built container image before comparison, see [`microsoft/integrity-vhd` CLI](https://github.com/microsoft/integrity-vhd/tree/main/cmd/dmverity-vhd).
+
 ## Reproduce measurements
 
-### Guest VM
+### 1. Verify security policy is the same
 
-Guest VM measurements can be authenticated using [platform endorsements](https://github.com/microsoft/confidential-aci-examples/blob/main/docs/Confidential_ACI_SCHEME.md#reference-info-base64). To get platform endorsements:
+`Host data` contains the hash of the security policy (e.g. `5ae7b14ee0c9c4fe267d191f25b20fffe24e29c4ac419c50501d20c869bbba65`). The policy can be obtained like it was shown above and saved to a file `ccepolicy.rego`. The hash of the Rego policy is the same as the one in the report:
```
achamayou, line 101:
```suggestion
`Host data` contains the sha-256 digest of the security policy (e.g. `5ae7b14ee0c9c4fe267d191f25b20fffe24e29c4ac419c50501d20c869bbba65`). The policy can be obtained like it was shown above and saved to a file `ccepolicy.rego`. The hash of the Rego policy is the same as the one in the report:
```

docs/reproducibility.md
```diff
@@ -8,48 +8,152 @@ The assumption here is that the original build was done using a Docker.
...
+UVM measurements can be authenticated using [platform endorsements](https://github.com/microsoft/confidential-aci-examples/blob/main/docs/Confidential_ACI_SCHEME.md#reference-info-base64). To get platform endorsements:
+
+```sh
+$ curl -s --cacert cacert.pem https://<LEDGER-URL>/node/quotes/self > node-quote.json
+$ cat node-quote.json | jq -r '.uvm_endorsements' | base64 -d > uvm_endorsements.cose
+... verify cose signing envelope ...
+```
 
-`Host data` contains the hash of the security policy, but we do not have the steps to obtain the policy yet. Container image would be linked to the policy.
+UVM endoresement policy can also be seen in `service-join-policy.json`.
```
achamayou, line 159:
```suggestion
UVM endoresment policy can also be seen in `service-join-policy.json`.
```

docs/reproducibility.md
ivarprudnikov: There are no instructions how to use hex2report and the last release was more than a year ago on Github releases. I would rather not confuse the reader.

docs/reproducibility.md
ivarprudnikov: The layers are in the policy, and this was exactly why I was interested in the policy. The fact that we have a join policy is secondary to the ability to get the policy to compare to the host_data measurement. If you are saying this is not guaranteed then there needs to be another API to get the security policy as we cannot regenerate it due to the complexity behind it.

docs/reproducibility.md
ivarprudnikov: The intention here is to reproduce the same layers that are in the policy. The policy seems to be keyed by its digest which in turn is in the report under Host data. Correct me if I am wrong but you need only one policy to inspect to get the layers.

docs/reproducibility.md
ivarprudnikov: added a note

docs/reproducibility.md
ivarprudnikov: added a note to use the host data measurement to get the policy

docs/reproducibility.md
ivarprudnikov: It is a good topic to discuss in a call, but let's release the changes to be able to have stable docker images and these steps that seem to work for now.

PR #273, opened 2025-02-21
Delete dpkg list files containing ppa.lauchpad.net from docker images

ivarprudnikov:
As part of the network isolation we've been instructed to get rid of ppa.lauchpad.net from the images.

It is introduced in CCF base images we use for running our application, e.g.:

```
~: $ docker run --rm -it ghcr.io/microsoft/ccf/app/run/virtual:ccf-6.0.0-dev19 /bin/bash
root@e957c9711c78:/# grep -Ril "ppa.launchpad.net" /etc
/etc/apt/sources.list.d/ppa_ubuntu_toolchain_r_test_focal.list
```


achamayou (approved):

achamayou:
CCF releases from 6.0.0-dev19 also include an Azure Linux RPM, which don't suffer from this problem.

Issue #272, opened 2025-02-20
Alignment with Feb 4 SCITT SCRAPI draft


ivarprudnikov:
Spec alignment with the specific [Feb 4 SCRAPI draft](https://github.com/ietf-wg-scitt/draft-ietf-scitt-scrapi/blob/4fafe531c21a3d1241934330bb771f82c64e0827/draft-ietf-scitt-scrapi.md) we have this API defined in the [Azure REST API specification using Typespec](https://github.com/Azure/azure-rest-api-specs/blob/a7358b9ca2ddab3e4794dcce6a25ef4c355dcaa8/specification/confidentialledger/Microsoft.CodeTransparency/main.tsp)

Similarly the ledger source needs to implement the missing bits to align with that specific SCRAPI draft:

**API changes**
- `GET .well-known/transparency-configuration` 
  - `jwksUri` needs to change to `jwks_uri`
  - in the case of an error it must be in CBOR format like in RFC draft
- `GET /jwks`
  - in the case of an error it must be in CBOR format like in RFC draft, making it JSON just because the response type is JSON would force to have duplicate login in the client dealing with the errors for both CBOR and JSON and the draft spec clearly says the errors should be CBOR
- `POST /entries`
  - in the case of an error it must be in CBOR format like in RFC draft
  - operation response must be in cbor
  - `OperationID` key must be capitalized like in the spec
  - `Status` key must be capitalized like in the spec
- `GET /operations/{operationId}`
  - operation response must be in cbor
  - errors must be in cbor
  - `EntryID` key must be capitalized like in the spec
  - `Status` key must be capitalized like in the spec
  - if the operation is still running then 202 response without body is returned
- `GET /entries/{entryId}`
  - errors must be in cbor
  - **response must return a cose receipt** and not a signed statement
- `GET /entries/{txid}/statement`
  -  errors must be in cbor

**Pyscitt changes** 
- following the API changes the Pyscitt implementation needs to align to newer output
- a new version of Pyscitt will need to be published

Removal of unused endpoints:
- `GET scitt/did.json` because we switch to jwks
- `GET .well-known/did.json` because we switch to jwks
- `GET /entries/{txid}/receipt` because 201 `/entries/{txid}` will return a receipt according to the spec

PR #274, opened 2025-02-21
Align with scrapi feb4

ivarprudnikov:
Following the details in the issue https://github.com/microsoft/scitt-ccf-ledger/issues/272

### DIFF
--- a/app/src/call_types.h
+++ b/app/src/call_types.h
@@ -2,9 +2,12 @@
 // Licensed under the MIT License.
 
 #pragma once
+#include "cbor.h"
 #include "kv_types.h"
+#include "odata_error.h"
 
 #include <ccf/ds/json.h>
+#include <ccf/tx_id.h>
 #include <nlohmann/json.hpp>
 #include <unordered_map>
 #include <vector>
@@ -81,5 +84,26 @@ namespace scitt
       std::optional<ODataError> error;
     };
   };
+
+  static std::vector<uint8_t> operation_to_cbor(
+    const GetOperation::Out& operation)
+  {
+    std::optional<std::string> entry_id_opt = operation.entry_id.has_value() ?
+      std::optional<std::string>(operation.entry_id.value().to_str()) :
+      std::nullopt;
+    std::optional<std::string> error_code_opt = operation.error.has_value() ?
+      std::optional<std::string>(operation.error.value().code) :
+      std::nullopt;
+    std::optional<std::string> error_message_opt = operation.error.has_value() ?
+      std::optional<std::string>(operation.error.value().message) :
+      std::nullopt;
+
+    return cbor::operation_props_to_cbor(
+      operation.operation_id.to_str(),
+      operationStatusToString(operation.status),
+      entry_id_opt,
+      error_code_opt,
+      error_message_opt);
+  }
 
 } // namespace scitt
--- a/app/src/call_types.h
+++ b/app/src/call_types.h
@@ -96,7 +96,7 @@ namespace scitt
   };
 
   DECLARE_JSON_TYPE(GetEntry::Out);
-  DECLARE_JSON_REQUIRED_FIELDS_WITH_RENAMES(GetEntry::Out, entry_id, "entryId");
+  DECLARE_JSON_REQUIRED_FIELDS_WITH_RENAMES(GetEntry::Out, entry_id, "EntryId");
 
   struct GetOperation
   {
@@ -111,9 +111,9 @@ namespace scitt
 
   DECLARE_JSON_TYPE_WITH_OPTIONAL_FIELDS(GetOperation::Out);
   DECLARE_JSON_REQUIRED_FIELDS_WITH_RENAMES(
-    GetOperation::Out, operation_id, "operationId", status, "status");
+    GetOperation::Out, operation_id, "OperationId", status, "Status");
   DECLARE_JSON_OPTIONAL_FIELDS_WITH_RENAMES(
-    GetOperation::Out, entry_id, "entryId", error, "error");
+    GetOperation::Out, entry_id, "EntryId", error, "Error");
 
   struct GetAllOperations
   {
--- a/app/src/call_types.h
+++ b/app/src/call_types.h
@@ -78,15 +78,8 @@ namespace scitt
       ccf::TxID operation_id;
       OperationStatus status;
       std::optional<ccf::TxID> entry_id;
-      // TODO: it needs to be in expcted structure as per RFC
-      std::optional<nlohmann::json> error;
+      std::optional<ODataError> error;
     };
   };
-
-  DECLARE_JSON_TYPE_WITH_OPTIONAL_FIELDS(GetOperation::Out);
-  DECLARE_JSON_REQUIRED_FIELDS_WITH_RENAMES(
-    GetOperation::Out, operation_id, "OperationId", status, "Status");
-  DECLARE_JSON_OPTIONAL_FIELDS_WITH_RENAMES(
-    GetOperation::Out, entry_id, "EntryId", error, "Error");
 
 } // namespace scitt
--- a/app/src/call_types.h
+++ b/app/src/call_types.h
@@ -11,22 +11,6 @@
 
 namespace scitt
 {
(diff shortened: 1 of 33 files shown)


app/src/http_error.h
```diff
@@ -3,57 +3,140 @@
...
     {}
+
+    /**
+     * Convert the error to a CBOR-encoded byte array.
+     * Follow rfc9290 for error encoding but use only
+     * title and detail and encode them cbor text.
+     */
+    std::vector<uint8_t> to_cbor_error() const
+    {
+      std::string error_mesasge = what();
```
achamayou, line 43:
```suggestion
      std::string error_message = what();
```

app/src/http_error.h
```diff
@@ -3,57 +3,140 @@
...
+      // otherwise decodign will fail
+      size_t buff_size = QCBOR_HEAD_BUFFER_SIZE + // map
+        QCBOR_HEAD_BUFFER_SIZE + // key
+        sizeof(cbor::CBOR_ERROR_TITLE) + // key
+        QCBOR_HEAD_BUFFER_SIZE + // value
+        code.size() + // value
+        QCBOR_HEAD_BUFFER_SIZE + // key
+        sizeof(cbor::CBOR_ERROR_DETAIL) + // key
+        QCBOR_HEAD_BUFFER_SIZE + // value
+        error_mesasge.size(); // value
```
achamayou, line 54:
```suggestion
        error_message.size(); // value
```

app/src/http_error.h
```diff
@@ -3,57 +3,140 @@
...
 #include "tracing.h"
 
+#include <qcbor/qcbor_decode.h>
+#include <qcbor/qcbor_encode.h>
+#include <qcbor/qcbor_spiffy_decode.h>
+#include <stdexcept>
+
 namespace scitt
 {
   struct HTTPError : public std::runtime_error
```
achamayou, line 16: Considering that all errors are CBOR in SCRAPI, I suspect we could s/HTTPError/HTTPCBORError/ and skip the flag altogether.

app/src/tracing.h
```diff
@@ -83,17 +83,19 @@ namespace scitt
...
-    const std::string& method,
+    const std::string& path,
     const std::function<ccf::ApiResult(::timespec& time)>& get_time,
     std::optional<ccf::TxID> txid = std::nullopt)
   {
     AppData& app_data = get_app_data(rpc_ctx);
 
     ::timespec end;
     ccf::ApiResult result = get_time(end);
+    // FIXME: should be CBOR or JSON
```
achamayou, line 94: Does that need to be done before merging, or is that separate?

app/src/tracing.h
```diff
@@ -147,44 +153,26 @@ namespace scitt
...
+        path,
+        ctx.rpc_ctx->get_request_query().c_str(),
         ctx.rpc_ctx->get_request_url());
 
       ::timespec start;
       ccf::ApiResult result = get_time(app_data.start_time);
       if (result != ccf::ApiResult::OK)
       {
-        SCITT_FAIL("Code=InternalError get_untrusted_host_time_v1 failed");
+        // FIXME: should be CBOR or JSON
```
achamayou, line 173: Does that need to be done before this merges, or is it separate?

pyscitt/pyscitt/client.py
```diff
@@ -22,6 +23,12 @@
 from .verify import ServiceParameters
 
 CCF_TX_ID_HEADER = "x-ms-ccf-transaction-id"
+CT_APPLICATION_JSON = "application/json"
+CT_APPLICATION_CBOR = "application/cbor"
+CT_APPLICATION_COSE = "application/cose"
+CT_APLICATION_CBOR_ERROR = "application/concise-problem-details+cbor"
```
achamayou, line 29:
```suggestion
CT_APPLICATION_CBOR_ERROR = "application/concise-problem-details+cbor"
```

pyscitt/pyscitt/client.py
```diff
@@ -277,6 +284,27 @@ def replace(self: SelfClient, **kwargs) -> SelfClient:
...
+        Parse the error response from the server and return a ServiceError instance.
+        """
+        if response.is_success:
+            return None
+
+        content_type = response.headers.get("content-type", CT_APPLICATION_JSON)
+        if (
+            content_type == CT_APPLICATION_CBOR
+            or content_type == CT_APPLICATION_COSE
+            or content_type == CT_APLICATION_CBOR_ERROR
```
achamayou, line 298:
```suggestion
            or content_type == CT_APPLICATION_CBOR_ERROR
```

test/test_historical.py
```diff
@@ -35,7 +35,7 @@ def test_enumerate_statements(self, client: Client, submissions):
                 start=submissions[0].seqno, end=submissions[-1].seqno
             )
         )
-
+        print("Found statements:", seqnos)
```
achamayou, line 38: Is this intended to merge?

app/src/service_endpoints.h
```diff
@@ -349,42 +250,22 @@ namespace scitt
...
-    registry
-      .make_endpoint(
-        "/scitt/did.json",
-        HTTP_GET,
-        ccf::json_adapter(std::bind(
-          endpoints::get_did_document, service_certificate_index, _1, _2)),
-        {ccf::empty_auth_policy})
-      .install();
-
+    // FIXME: /jwks should return CBOR errors by default
```
achamayou, line 253: This is out of the scope of the draft, but I think it'd be nice to provide a CBOR endpoint for this too/instead, and remove that last bit of JSON. There is no CWK though, but there is a COSE Key format: https://www.rfc-editor.org/rfc/rfc8152#page-33

I think the right way to do this is to have an API to get the current ledger CoseKey from CCF (same as the service cert), and to expose that at a `/cose_keys` or similar, and update the transparency config accordingly. This is very much for later though and not something that requires short term attention.

app/src/tracing.h
ivarprudnikov: Technically speaking it should be done now

app/src/tracing.h
ivarprudnikov: As mentioned in another comment it could in theory fail on SCITT endpoints

app/src/http_error.h
ivarprudnikov: These are managed errors that we throw. In some cases our API would return whatever CCF returns and it would make things slightly more awkward if our errors were all CBOR but some other errors were JSON. Now at least the JSON endpoints return JSON errors and CBOR endpoints return CBOR. Some native support of `ser_error` in CCF would be great.

achamayou: That's definitely something we ought to look into, yes.

app/src/tracing.h
ivarprudnikov: Replaced with explicit CBOR error

app/src/tracing.h
ivarprudnikov: Replaced with explicit CBOR error

app/src/call_types.h
```diff
@@ -87,55 +71,22 @@ namespace scitt
...
-
   struct GetOperation
   {
     struct Out
     {
       ccf::TxID operation_id;
       OperationStatus status;
       std::optional<ccf::TxID> entry_id;
+      // TODO: it needs to be in expcted structure as per RFC
       std::optional<nlohmann::json> error;
```
ivarprudnikov, line 82: @achamayou Is there some sane way to express this error in a structure which will be then converted to cbor?

At the moment the implementation does create a struct then converts to json then json to cbor:
```
operation = {
.operation_id = operation_id,
.status = OperationStatus::Failed,
.error =
  ODataError{
    .code = ccf::errors::TransactionInvalid,
    .message = "Transaction is invalid",
  },
};
...
auto body = nlohmann::json(operation);
ctx.rpc_ctx->set_response_body(nlohmann::json::to_cbor(body));
```

app/src/http_error.h
ivarprudnikov: I stuck with having 2 sets of errors, we can revisit later and remove JSON altogether

PR #277, opened 2025-02-24
Consolidate on IETF SCITT profile/policyScript

achamayou:
`CWT_Claims` are mandatory in the [Architecture, Section 4.2](https://www.ietf.org/archive/id/draft-ietf-scitt-architecture-10.html#name-signed-statements):

> The protected header of a Signed Statement and a Receipt MUST include the CWT Claims header parameter as specified in [Section 2](https://datatracker.ietf.org/doc/html/draft-ietf-cose-cwt-claims-in-headers-10#section-2) of [[CWT_CLAIMS_COSE](https://www.ietf.org/archive/id/draft-ietf-scitt-architecture-10.html#CWT_CLAIMS_COSE)]. The CWT Claims value MUST include the Issuer Claim (Claim label 1) and the Subject Claim (Claim label 2) [[IANA.cwt](https://www.ietf.org/archive/id/draft-ietf-scitt-architecture-10.html#IANA.cwt)].

Signing services that are going to use CTS have agreed to supply it (separate email thread), and so the code can be simplified accordingly. Statements without these fields are not realistically useful to relying parties anyway, who would need to parse the `x5chain` to establish a stable issuer identity.

In addition, this removes the duplicate and less flexible CA store functionality (also referenced in the separate email thread), in favour of policy. A clarification was added to the [Architecture document](https://github.com/ietf-wg-scitt/draft-ietf-scitt-architecture/pull/345), as suggested by the signing services team.


achamayou:
Closed in favour of branch version (#278).

PR #278, opened 2025-02-25
Consolidate on IETF SCITT profile/policyScript

achamayou:
`CWT_Claims` are mandatory in the [Architecture, Section 4.2](https://www.ietf.org/archive/id/draft-ietf-scitt-architecture-10.html#name-signed-statements):

> The protected header of a Signed Statement and a Receipt MUST include the CWT Claims header parameter as specified in [Section 2](https://datatracker.ietf.org/doc/html/draft-ietf-cose-cwt-claims-in-headers-10#section-2) of [[CWT_CLAIMS_COSE](https://www.ietf.org/archive/id/draft-ietf-scitt-architecture-10.html#CWT_CLAIMS_COSE)]. The CWT Claims value MUST include the Issuer Claim (Claim label 1) and the Subject Claim (Claim label 2) [[IANA.cwt](https://www.ietf.org/archive/id/draft-ietf-scitt-architecture-10.html#IANA.cwt)].

Signing services that are going to use CTS have agreed to supply it (separate email thread), and so the code can be simplified accordingly. Statements without these fields are not realistically useful to relying parties anyway, who would need to parse the `x5chain` to establish a stable issuer identity.

In addition, this removes the duplicate and less flexible CA store functionality (also referenced in the separate email thread), in favour of policy. A clarification was added to the [Architecture document](https://github.com/ietf-wg-scitt/draft-ietf-scitt-architecture/pull/345), as suggested by the signing services team.

Branch version of #277


docs/configuration.md
```diff
@@ -113,7 +113,7 @@ Policy scripts are executed by the [CCF JavaScript runtime](https://github.com/m
 Example `set_scitt_configuration` snippet:
 ```json
 "policy": {
-  "policyScript": "export function apply(profile, phdr) {\n  if (profile === \"X509\") { return true; }\n  return \"Only X509 claim profile is allowed\";\n}"
+  "policyScript": "export function apply (phdr) { return true; }"
```
andpiccione, line 116: Please remove the `Trust stores` section below as users would no longer need to configure x509 roots.

pyscitt/pyscitt/governance.py
```diff
@@ -157,17 +157,6 @@ def set_scitt_configuration_proposal(configuration: dict) -> dict:
     }
 
 
-def set_ca_bundle_proposal(name: str, bundle: str) -> dict:
```
andpiccione, line 160: Please update the demo scripts accordingly as we no longer need to set the x509 roots of trust for the instance configuration.

achamayou: I think the scripts themselves need updating to set a CWT claims and generate a policy. If we want the demo to continue to work, I suggest we find a way to hook it up to the CI, otherwise it's pretty much guaranteed to go stale.

test/infra/fixtures.py
ivarprudnikov, line 1: `trusted_ca` needs review as it was supposed to be used to create cose signatures but those should fail now that we do not have x509 roots configured in the tests. It either needs to change to something that sets the config with the expected policy that matches the generated key or should be superseded by `configure_service`.

test/test_cli.py
```diff
@@ -59,176 +59,6 @@ def test_smoke_test(run, client, tmp_path: Path):
     )
 
 
-def test_use_cacert_submit_verify_x509_signature(run, client, tmp_path: Path):
```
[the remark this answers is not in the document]
achamayou: I think we test end to end flows in test_ccf and test_configuration already, but I am always in favour of expanding test coverage.

ivarprudnikov (approved):

PR #288, opened 2025-03-06
update demo readme and scripts and minor changes in pyscitt client and test infra module

ivarprudnikov:
- pyscitt cli fixes in pretty receipt logic to make sure it works with transparent statements
- test infra generate script operation simplification and expanded support to add eku to end entity cert
- explicit config template in demo and expected issuer string creation
- demo readme updates
- demo script updates


andpiccione (approved):

DEVELOPMENT.md
```diff
@@ -105,7 +105,7 @@ The application expects the [configuration](docs/configuration.md) to be submitt
 echo <<< EOL
 {
     "policy": {
-        "policyScript": "export function apply(phdr) { if (!phdr.issuer) {return 'Issuer not found'} if (phdr.issuer !== 'did:x509:0:sha256:HnwZ4lezuxq/GVcl/Sk7YWW170qAD0DZBLXilXet0jg=::eku:1.3.6.1.4.1.311.10.3.13') { return 'Invalid issuer'; } }"
+        "policyScript": "export function apply(phdr) { if (!phdr.issuer) {return 'Issuer not found'} else if (phdr.issuer !== 'did:x509:0:sha256:HnwZ4lezuxq/GVcl/Sk7YWW170qAD0DZBLXilXet0jg=::eku:1.3.6.1.4.1.311.10.3.13') { return 'Invalid issuer'; } return true; }"
```
achamayou, line 108: I was thinking more of a line return, and a series of if/early returns here, but this works too of course.

pyscitt/pyscitt/cli/pretty_receipt.py
```diff
@@ -14,16 +14,16 @@
 
 def prettyprint_receipt(receipt_path: Path):
     """
-    Pretty-print a COSE receipt file
+    Pretty-print a receipt which could be a standalone cose file or a transparent statement.
```
achamayou, line 17: I think that's fine to do, but then let's change the name from pretty-receipt to pretty-cose, because Statements are not Receipts (they can contain Receipts, but don't have to).

achamayou (approved):

DEVELOPMENT.md
ivarprudnikov: it was failing as it was for me because there was no return in the case of success

pyscitt/pyscitt/cli/pretty_receipt.py
ivarprudnikov: Some people think receipt is a transparent statement and some that it is a standalone receipt. This is effectively to make sure it prints even if user provides a transparent statement by accident

PR #289, opened 2025-03-10
Removed Python dependencies step

vimauro:
- Removed python.yml since it's not needed anymore. We are now referencing directly only ado_pipauth.yml.
- Explicitly delete any reference to **launchpad** in the repository files

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -63,10 +63,10 @@ jobs:
       # Manually disable Launchpad repository
       - name: Disable Launchpad and PPA repositories
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test.yml
```diff
@@ -59,6 +59,15 @@ jobs:
       # Work-around for https://github.com/actions/runner/issues/2033
       - name: Work around git safe.directory in container
         run: chown -R $(id -u):$(id -g) $PWD
+
+      # Manually disable Launchpad repository
```
ivarprudnikov, line 63:
```suggestion
      # Manually disable Launchpad repository to reduce the chances of a supply chain attack
```

PR #290, opened 2025-03-11
Remove usage of CCF private headers

achamayou:
Depends on CCF 6.0.0-rc1 being released (https://github.com/microsoft/CCF/pull/6895), and scitt-ccf-ledger being updated accordingly.


maxtropets:
Moved all the changes to #291 as that won't build without it (include path mismatch due to openssl wrapper move in 6.0.0-rc1).

PR #291, opened 2025-03-12
Update to CCF 6.0.0-rc1

maxtropets:


ivarprudnikov:
@maxtropets the PR should not be done from the fork, Azure pipelines cannot run in that case, given that you have access then create a branch directly in this repo

PR #292, opened 2025-03-12
Update to CCF 6.0.0-rc1

maxtropets:
Same as #291, but created from microsoft remote instead of a fork.

PR #298, opened 2025-04-25
Update to CCF 6.0.1 and use Azure Linux 3.0

andpiccione:
Starting with version 6.0.0, CCF stopped publishing runtime and development container images and targets Azure Linux 3.0 only. This PR updates all the development scripts and pipelines to run on AZL3 and use the new CCF RPM packages.

The PR contains the following changes:

- Update devcontainer to run on AZL3 base image and with all the required dev dependencies
- Update ADO pipeline and GitHub Actions to run on 1ES hosted runners based on AZL3 image
- Update DockerFile to consume new CCF dependencies and install the minimal packages required at runtime (this also results in a 3x smaller container image!)
- Update clang, clang-tidy, and clang-format to version 18 and format code
- Remove old python 3.8 and 3.10 references and rely on python3 available in AZL3 images
- Remove all references of old CCF images, Ubuntu commands, Debian packages, etc.
- Add new stage to PR build to push container images to staging ACR (useful for testing purposes)


scripts/setup-env.sh
```diff
@@ -0,0 +1,34 @@
...
+
+tdnf install -y \
+    git \
+    build-essential \
+    python3-pip \
+    ca-certificates \
+    jq \
+    which \
+    procps \
+    clang-tools-extra-devel
```
ivarprudnikov, line 20: please use specific versions to make it transparent what is installed and to make it reproducible

.github/workflows/build-test.yml
```diff
@@ -60,15 +70,7 @@ jobs:
       - name: Work around git safe.directory in container
         run: chown -R $(id -u):$(id -g) $PWD
 
-      # Manually disable Launchpad repository to reduce the chances of a supply chain attack
-      - name: Disable Launchpad and PPA repositories
-        run: |
-          sudo sed -i '/launchpad.net/d' /etc/apt/sources.list
-          sudo sed -i '/ppa.launchpad.net/d' /etc/apt/sources.list
-          sudo find /etc/apt/sources.list.d/ -type f -exec sed -i '/launchpad.net/d' {} \;
-          sudo find /etc/apt/sources.list.d/ -type f -exec sed -i '/ppa.launchpad.net/d' {} \;
```
ivarprudnikov, line 69: Is it not an issue anymore, previously those were part of the image

docker/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+# Use fixed image version to ensure reproducibility
+ARG BASE_IMAGE=mcr.microsoft.com/azurelinux/base/core:3.0.20250402
+
+FROM ${BASE_IMAGE} AS builder
+
+ARG CCF_VERSION
+ARG CCF_PLATFORM
+ARG SCITT_VERSION_OVERRIDE
+
+RUN tdnf install -y ca-certificates && \
```
ivarprudnikov, line 12: please use specific versions to aid in reproducibility

docker/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+ARG CCF_VERSION
+ARG CCF_PLATFORM
+
+WORKDIR /usr/src/app
+
+RUN tdnf install -y \
+    ca-certificates \
+    tini \
+    tar \
+    nc && \
```
ivarprudnikov, line 43: please use specific versions to aid in reproducibility

docker/run-dev.sh
```diff
@@ -104,16 +104,17 @@ docker run --name "$CONTAINER_NAME" \
...
+    --entrypoint "cchost" \
     "$DOCKER_TAG" --config /host/dev-config.json
 
 echo "Setting up python virtual environment."
 if [ ! -f "venv/bin/activate" ]; then
-    if command -v python3.10 &> /dev/null; then
-        PYTHON=python3.10
-    elif command -v python3.8 &> /dev/null; then
-        PYTHON=python3.8
+    if command -v python &> /dev/null; then
```
ivarprudnikov, line 112: I think we want to be specific as to which python version is being used, i.e. pyscitt is said to be compatible with python3.8+ so either we increase that number or we make sure we test with it to avoid some weird regressions.

start.sh
```diff
@@ -24,14 +24,14 @@ fi
 
 echo "Setting up python virtual environment."
 if [ ! -f "venv/bin/activate" ]; then
-    python3.8 -m venv "venv"
+    python3 -m venv "venv"
```
ivarprudnikov, line 27: the version is explicit is because we say we support it in pyscitt config, why drop it?

scripts/setup-env.sh
ivarprudnikov: IMO this is not as important as dockerfile

pyscitt/setup.py
```diff
@@ -25,7 +25,7 @@
     },
     python_requires=">=3.8",
     install_requires=[
-        "ccf==6.0.0-rc0",
+        "ccf==6.0.1",
```
ivarprudnikov, line 28: package version should also be increased

andpiccione: Addressed in https://github.com/microsoft/scitt-ccf-ledger/pull/298/commits/7fe186fc89fd4f3d28f61b555cf96a1f05acc0b3

docker/Dockerfile
andpiccione: Addressed in https://github.com/microsoft/scitt-ccf-ledger/pull/298/commits/bf22322a96a07fda8eae637247b494145222c10c

docker/Dockerfile
andpiccione: Addressed in https://github.com/microsoft/scitt-ccf-ledger/pull/298/commits/bf22322a96a07fda8eae637247b494145222c10c

.github/workflows/build-test.yml
andpiccione: Launchpad and PPA repositories were used for Ubuntu / Debian packages. This should no longer be a problem since Azure Linux 3 uses RPM as a package manager.

.pipelines/pullrequest.yml
```diff
@@ -176,3 +189,33 @@ extends:
             displayName: Build snp with Docker
             env:
               PLATFORM: snp
+
+    - stage: docker
```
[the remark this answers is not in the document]
ivarprudnikov: Basically my issue with this is that we then need to maintain the identities and use ev2 to push those images, and that part was not elegant to deal with, not to mention s360 telling us to change things. I would argue we should not open a possibility to risk using the PR built image in production anyway. Testing in clusters is usually carried out after the changes get merged to main, these PR images will be left unused anyway. Another thing against it is that if you have many changes to the PR many images will be rebuilt and pushed (I know the layers might be reused anyway). Less is more in my view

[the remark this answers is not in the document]
andpiccione: > we then need to maintain the identities

We still need the same identities in the Official build pipeline, so we still have to maintain them anyway.

> use ev2 to push those images

After the recent refactoring, we no longer use ev2 to push images; we push directly from the build pipeline.

> we should not open a possibility to risk using the PR built image in production anyway

I agree, and in fact we are not even pushing the images to a production ACR. The current implementation only pushes the images to a staging ACR for testing, so no production touch at all. Pushing to staging is enough for testing.

> Testing in clusters is usually carried out after the changes get merged to main, these PR images will be left unused anyway

This is usually true, but there have been instances in the past where we needed to test an image before merging the PR, such as for SNP-specific tasks. Even taking the example of this PR, we definitely want to quickly run a ledger using this image to validate the Azure Linux 3.0 support before merging thse changes, to avoid the risk we have to make patches after creating a new SCITT release.

> Another thing against it is that if you have many changes to the PR many images will be rebuilt and pushed (I know the layers might be reused anyway). Less is more in my view

I agree, even though we use the same approach for all our other internal services and it hasn't caused any problems so far. Perhaps we can make the new stage optional: by default, we won't build and push the image from PR builds unless we enable a specific flag when manually creating a PR build. What do you think?

demo/cts_poc/0-cacerts-generator.sh
```diff
@@ -14,7 +14,7 @@ CUSTOM_EKU="1.3.6.1.5.5.7.3.36"
 
 echo -e "\nSetting up environment"
 if [ ! -f "venv/bin/activate" ]; then
-    python3.8 -m venv "venv"
+    python3.12 -m venv "venv"
```
achamayou, line 17: I believe the 3.8 is from the days of using old Ubuntu, and installing newer python through the deadsnakes ppa, and python3 still pointing to the older default. On Azure Linux 3.0, 3 is 3.12, which is very recent, and so I think we could probably revert to just `python3` everywhere.

That will cause a little less friction if Azure Linux updates python3, or when we update to Azure Linux 4.0, and should not create any problems.

It's a very minor thing obviously.

achamayou (approved):

achamayou:
Supersedes #292

achamayou:
I am unable to find the page at the moment, but I believe I have a seen documentation that showed how to pin all the versions in an Azure Linux image in one go, and I think that's what we want to do for reproducibility.

demo/cts_poc/0-cacerts-generator.sh
andpiccione: That's what I initially did, but Ivar pointed out [above](https://github.com/microsoft/scitt-ccf-ledger/pull/298#discussion_r2063986070) that we want to pin to a specific Python version since we say pyscitt is compatible with certain Python versions only. If the python3 binary packaged with AZL3 changes and points to a new Python version (e.g., python 3.13), we risk losing control over which version is supported in pyscitt.

andpiccione:
> I am unable to find the page at the moment, but I believe I have a seen documentation that showed how to pin all the versions in an Azure Linux image in one go, and I think that's what we want to do for reproducibility.

If it's possible to pin the packages versions in the Dockerfile in a simpler way, that would be great!

.pipelines/pullrequest.yml
```diff
@@ -176,3 +189,33 @@ extends:
...
+      jobs:
+      # Build one image per each platform and push it to ACR
+      
+      # Virtual
+      - template: .pipelines/obpdocker.yml@self
+        parameters:
+          Name: scitt_virtual
+          DisplayName: Build Virtual Docker image
+          AcrName: 'confidentialledgeracrstaging'
+          DockerTag: $(gitTagOrCommit)_$(Build.BuildNumber)
```
ivarprudnikov, line 207: Please update the tags to make sure that the image is from a PR, e.g. `$(gitTagOrCommit)_$(Build.BuildNumber)_pr` if you still need to push the images.

ivarprudnikov (approved):

.pipelines/pullrequest.yml
andpiccione: Addressed in https://github.com/microsoft/scitt-ccf-ledger/pull/298/commits/8d8590b86f438ebe81b05d19f7cbc880437c55a6

ivarprudnikov (approved):

PR #305, opened 2025-07-03
Feat/tss msft css dev attestation verification

ivarprudnikov:
Building on top of #297 but a bit more fine grained:
- Authenticate new issuer type `did:attestedsvc:msft-css-dev`
- Use protected headers

    ```
    { 
       1: -35,                ; alg: ES384 
       2: [ "msft-css-dev" ], ; crit 
       4: bstr,               ; kid: SHA256 hash of canonical COSE_Key 
      15: { ... },            ; CWT-Claims including issuer/subject 
      "msft-css-dev": { ... } ; attestation map
    }
    ```

- Verify attestation map provided in `msft-css-dev`:

    ```
    { 
      "attestation": bstr,       ; raw hardware attestation report 
      "attestation_type": tstr,  ; "SEV-SNP:ContainerPlat-AMD-UVM" 
      "cose_key": { ... },       ; canonical COSE_Key map (embedded directly) 
      "snp_endorsements": bstr,  ; concatenated PEM-encoded AMD cert chain 
      "uvm_endorsements": bstr,  ; opaque endorsement from Azure or UVM authority 
      "ver": int                 ; version number of the format (e.g., 0) 
    }
    ```

    - Verification steps:

    1. Recognize the signature scheme based on the beginning of the issuer DID: did:attestedsvc 
    2. Confirm the attested service is recognized and supported, i.e., did:attestedsvc:msft-css-dev::...  
    3. Validate the attestation in the corresponding protected header against the AMD certificate chain contained in “snp_endorsements”. 
    4. Enforce policy: 
        - Container measurement matches the expected configuration for C-SS 
        - Report_data matches the hash of the provided COSE_Key 
    5. Extract the COSE_Key from the attestation and ensure its hash matches the signature keyId. 
    6. Verify the COSE signature using the COSE_Key.

- Expose the values for the policy


.github/workflows/codeql.yml
```diff
@@ -78,7 +78,7 @@ jobs:
     - if: matrix.language == 'cpp'
       name: Build C++
       run: |
-        CMAKE_BUILD_TYPE=Debug PLATFORM=virtual BUILD_TESTS=OFF ./build.sh
```
achamayou, line 81: From a quick look at build.sh, the default seems to have changed to Debug, and so the behaviour is still the same I guess?

.github/workflows/long-test.yml
```diff
@@ -37,11 +37,6 @@ jobs:
         run: |
           ./scripts/setup-env.sh
           ./build.sh
-
-      - name: Unit tests
```
achamayou, line 41: This ran the unit tests with ASAN on, which seems useful?

app/src/cbor.h
```diff
@@ -153,4 +153,39 @@ namespace scitt::cbor
     output.shrink_to_fit();
     return output;
   }
+
+  // see https://www.ietf.org/rfc/rfc9679.html
+  inline std::vector<uint8_t> cose_key_to_cbor(
```
achamayou, line 158:
```suggestion
  inline std::vector<uint8_t> ec_cose_key_to_cbor(
```
other kty have different params.

app/src/cbor.h
```diff
@@ -153,4 +153,39 @@ namespace scitt::cbor
     output.shrink_to_fit();
     return output;
   }
+
+  // see https://www.ietf.org/rfc/rfc9679.html
```
achamayou, line 157:
```suggestion
  // see https://www.ietf.org/rfc/rfc9679.html#section-4.2
```

app/src/cose.h
```diff
@@ -60,11 +81,156 @@ namespace scitt::cose
 
   static constexpr const char* SVN_HEADER_PARAM = "svn";
 
+  static std::shared_ptr<ccf::crypto::HashProvider>& get_hash_provider()
+  {
+    static thread_local std::shared_ptr<ccf::crypto::HashProvider>
```
achamayou, line 86: I am surprised that this is worth caching, because:

1. HashProvider (OpenSSLHashProvider really) construction is trivial, there is nothing to initialise
2. There are unfortunate lookup costs per call (not per construction) in OpenSSL3, but caching the provider does not help with those and they are mitigated by CCF already (https://github.com/microsoft/CCF/blob/main/src/crypto/openssl/hash.cpp#L59)

app/src/cose.h
```diff
@@ -60,11 +81,156 @@ namespace scitt::cose
...
+  }
+
   struct COSEDecodeError : public std::runtime_error
   {
     COSEDecodeError(const std::string& msg) : std::runtime_error(msg) {}
   };
 
+  // Cose Key https://www.rfc-editor.org/rfc/rfc9679.html
+  // Presence of values depends on kty
+  class CoseKeyMap
```
achamayou, line 106: Arguably ECCoseKeyMap in its current form.

app/src/policy_engine.h
```diff
@@ -123,6 +122,94 @@ namespace scitt
...
+            cose_key_obj.set(
+              "y", ctx.new_array_buffer_copy(cose_key.y().value()));
+          }
+
+          tss_map.set("cose_key", std::move(cose_key_obj));
+
+          auto cose_key_sha256 = cose_key.to_sha256_thumb();
+          tss_map.set(
+            "cose_key_sha256",
+            ctx.new_string(ccf::crypto::Sha256Hash(cose_key_sha256).hex_str()));
```
achamayou, line 172: This will digest again, you want something like from_span() to set the value directly.

app/src/verifier.h
```diff
@@ -173,36 +178,172 @@ namespace scitt::verifier
...
+      quote_info.endorsements = phdr.tss_map.snp_endorsements.value();
+
+      ccf::pal::PlatformAttestationMeasurement measurement = {};
+      ccf::pal::PlatformAttestationReportData report_data = {};
+      std::optional<ccf::pal::UVMEndorsements> parsed_uvm_endorsements;
+
+      try
+      {
+        ccf::pal::verify_snp_attestation_report(
+          quote_info, measurement, report_data);
```
achamayou, line 248: This extracts the measurement from the attestation, but I think we want to go further than this and call `ccf::pal::verify_uvm_endorsements_descriptor()` with the endorsements to extract a descriptor `(did, feed, svn) atm, CWT claims later`, which can be exposed to the policy.
It's fine to expose the raw measurement too (why not), but policies that check it directly will be quite brittle.

app/src/verifier.h
```diff
@@ -173,36 +178,172 @@ namespace scitt::verifier
...
+      }
+      catch (const std::exception& e)
+      {
+        throw VerificationError(
+          fmt::format("Failed to compute COSE key hash: {}", e.what()));
+      }
+
+      // select first 32 bytes of the report data
+      std::span<uint8_t> report_data_span(
+        report_data.data.data(), std::min(report_data.data.size(), size_t(32)));
```
achamayou, line 278: It would be a bit stricter and therefore preferable to make sure the rest is zeros.

app/unit-tests/cose_test.cpp
```diff
@@ -22,7 +22,123 @@ using namespace testutils;
...
+  // add a test case to use payloads from test/payloads directory
+  // NOLINTBEGIN(bugprone-unchecked-optional-access)
+  TEST(CoseTest, DecodeTSSHeaders)
+  {
+    std::string filepath = "test_payloads/css-attested-cosesign1-20250617.cose";
+    std::ifstream file(filepath, std::ios::binary);
+    ASSERT_TRUE(file.is_open());
+
+    // Get file size
+    file.seekg(0, std::ios::end);
```
achamayou, line 43: There's a more readable `std::filesystem::file_size()`, but it takes a path and so leads to a double open, ymmv.

build.sh
```diff
@@ -44,31 +38,7 @@ if [ "$BUILD_CCF_FROM_SOURCE" = "ON" ]; then
...
-    -DENABLE_CLANG_TIDY="${ENABLE_CLANG_TIDY}" \
-    "$root_dir/app"
-
-ninja -C build/app ${NINJA_FLAGS} --verbose
-ninja -C build/app ${NINJA_FLAGS} install
-
-echo "List of installed files in SCITT_DIR $install_dir"
-ls -R $install_dir
+cd app
+cmake --workflow --preset "$PLATFORM" --fresh
```
achamayou, line 42: Ok I see, this changes the default build from Release to Debug, which is fine, but I guess it's important to check the release pipelines are all setting this explicitly and not getting debug builds. Also the BUILD_CCF_FROM_SOURCE is doing Debug/ASAN, which is fine, but can maybe be renamed to BUILD_DEBUG_CCF_FROM_SOURCE to avoid accidents. Quite separate from the main goal of the PR, but since this change is happening here, I thought I'd mention it anyway :)

pyscitt/pyscitt/client.py
```diff
@@ -231,7 +231,7 @@ def __init__(
             A bearer token for all requests made by this instance.
 
         member_auth:
-            MemberAuthenticationMethod include A pair of certificate and private key in PEM format or AKV login identity, used to sign requests.
+            MemberAuthenticationMethod include A pair of certificate and private key in PEM format, used to sign requests.
```
achamayou, line 234: Dropping this seems fine, but could probably be a separate PR without drawback and with a slight clarity benefit?

app/src/cose.h
ivarprudnikov: This was done to be able to do unit tests as they would otherwise fail due to init not called. And init seems to be not exported in a way to be invoked in a test, hence this method :(

build.sh
ivarprudnikov: when releasing we build in docker containers without invoking this script, otherwise the var renamed

app/src/cose.h
achamayou: Ah, I see! Yes, it's not in any exported headers at the moment, but that's something we could fix easily.

.github/workflows/codeql.yml
ivarprudnikov: it is all the same I just moved variables to a preset as it is being used all over the place over and over again, not to mention that prod builds happen in a container anyway and this is more of a development/debug/ci script

app/unit-tests/cose_test.cpp
ivarprudnikov: yep it is easier to read

app/src/cose.h
```diff
@@ -60,11 +81,156 @@ namespace scitt::cose
...
+    void set_x_e(std::optional<std::vector<uint8_t>> x_e)
+    {
+      x_e_ = x_e;
+    }
+    void set_y(std::optional<std::vector<uint8_t>> y)
+    {
+      y_ = y;
+    }
+
+    void validate() const
```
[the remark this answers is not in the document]
ivarprudnikov: when parsing cbor, setters are used instead of ctor so moving it to ctor will not solve the issue

app/src/cbor.h
ivarprudnikov: separated into ec_cose_key_to_cbor and rsa_cose_key_to_cbor and added  basic tests

app/src/cose.h
ivarprudnikov: added support for RSA, some explanation in comments

achamayou (approved):
LGTM as an initial step, but I think we do want to verify the UVM endorsement before the policy, and expose the descriptor (did, feed, svn for now, later the scitt compliant CWT_Claims) to be able to write sustainable policies.

Having said that, this already allows end to end testing with policy-hardcoded UVM measurements, and that's very useful to move forward.

app/src/policy_engine.h
```diff
@@ -151,10 +278,13 @@ namespace scitt
...
 
       auto phdr_val = protected_header_to_js_val(interpreter, phdr);
+      auto uhdr_val = unprotected_header_to_js_val(interpreter, uhdr);
+      auto payload_val = interpreter.new_array_buffer_copy(payload);
+      auto details_val = verified_details_to_js_val(interpreter, details);
 
       const auto result = interpreter.call_with_rt_options(
         apply_func,
-        {phdr_val},
+        {phdr_val, uhdr_val, payload_val, details_val},
```
andpiccione, line 287: I suppose this won't break for existing registration policies that only define a single argument for the protected header, right? In other words, the other arguments will be just ignored, correct?

andpiccione, line 287: Because we are now passing the entire payload to the JS interpreter, do we need to increase the max_stack_bytes limit set below? Since the entire payload can go up to 1MB, I suppose the stack should be higher than that to accommodate for large parameters.

pyscitt/pyscitt/cli/client_arguments.py
```diff
@@ -76,11 +74,6 @@ def add_client_arguments(
             type=Path,
             help="Path to CCF member certificate file",
         )
-        parser.add_argument(
```
andpiccione, line 79: Are we removing this because it's no longer used/needed? I think it was added in the past to support the signing of claims using a key stored in AKV (I vaguely recall it was asked by some customer). Feel free to ignore the comment if that's the case.

andpiccione (approved):

app/src/policy_engine.h
ivarprudnikov: Yes it will not affect the existing policies. We do have a couple of configuration tests where arguments vary.

pyscitt/pyscitt/cli/client_arguments.py
ivarprudnikov: Yeah we do not use it at all and it is not being tested to understand if it even works anymore. This is also mainly used in testing as pyscitt _hides_ signing command, e.g.:

```
./pyscitt.sh --help
usage: scitt [-h] {submit,retrieve,pretty-receipt,embed-receipt,validate,split-payload} ...

positional arguments:
  {submit,retrieve,pretty-receipt,embed-receipt,validate,split-payload}
                        Choose one of the available commands to run. Use the --help flag to see the options for each command. For instance 'scitt submit --help' will show the options for the submit command.

options:
  -h, --help            show this help message and exit
```

^^^ no `sign` command in list. This is to make sure we do not need to maintain complicated signing mechanisms and think about backward compatibility too much.

andpiccione (approved):

PR #326, opened 2025-09-25
Draft re-shape of the attestedsvc map

achamayou:
```
{
   1: -35,                ; alg: ES384
   2: [ "attestedsvc" ], ; crit
   4: bstr,               ; kid: SHA256 hash of canonical COSE_Key
  15: { ... },            ; CWT-Claims including issuer/subject
  "attestedsvc": { "svc_id": "msft-css-dev", ... } ; attestation map as before
}
```

@briankr-ms this is `svc_id` rather than `svc-id` for consistency with the other fields in the map (`uvm_endorsement` etc). I am not clear on the utility, other than it being a minor convenience for the policy writer, since this is endorsed by the same key and under the same circumstances as Issuer.

TODO:

- [ ] Tests
- [ ] Confirm crit verification works
- [ ] Do we need `svc_id`? It seems redundant with the one in the issuer, and is equally authoritative.

### DIFF
--- a/app/src/cose.h
+++ b/app/src/cose.h
@@ -37,7 +37,8 @@ namespace scitt::cose
   static constexpr int64_t COSE_HEADER_PARAM_X5CHAIN = 33;
   static constexpr int64_t COSE_HEADER_PARAM_CWT_CLAIMS = 15;
 
-  static constexpr const char* COSE_HEADER_PARAM_TSS = "msft-css-dev";
+  static constexpr const char* COSE_HEADER_PARAM_TSS = "attestedsvc";
+  static constexpr const char* COSE_HEADER_PARAM_TSS_SVC_ID = "svc_id";
   static constexpr const char* COSE_HEADER_PARAM_TSS_ATTESTATION =
     "attestation";
   static constexpr const char* COSE_HEADER_PARAM_TSS_ATTESTATION_TYPE =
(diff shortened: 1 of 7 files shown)


test/test_configuration.py
```diff
@@ -406,11 +406,13 @@ def didx509_signed_statement_with_attestation(
             {"foo": "bar"},
             cwt=True,
             additional_phdr={
-                "msft-css-dev": {
+                2: ["attested_svc"],
```
ivarprudnikov, line 409: incorrect crit header value

ivarprudnikov: to effect this test should be failing

achamayou: That is interesting, I briefly thought this was incompatible with the PR intent, but it's actually not. The change is to check that crit is set as we expect when we validate "did:attestedsvc:" issuer payloads (see https://github.com/microsoft/scitt-ccf-ledger/pull/326/files#diff-e1a224cc1332840ca47edea8a0285dd6606e707da3097bdc4e6b3007ebd01b95L369 for the code path).

There is no crit check for other issuer types, including did:x509, and so didx509_signed_statement_with_attestation is accepted despite the mistyped crit.

I see two follow up possibilities:

1. remove the use of crit in this test, which is confusing
2. additionally enforce that crit is empty when using did:x509

I suggest that 2. is best done in its own PR, with additional tests. I also want to check if we should have `x5chain` in crit for did:x509, I know that the spec says that at least core COSE values should never be there, but I am not sure about `x5chain`.

ivarprudnikov (approved):

PR #260, opened 2025-01-30
Draft: Rego policy support

achamayou:
Depends on #259, needs samples and documentation

PR #338, opened 2025-10-23
Add support for Rego registration policies (rego-cpp 1.2.0)

achamayou:
Add support for a "policyRego" field in configuration, similar to "policyScript" but allowing the definition of [Rego](https://www.openpolicyagent.org/docs/latest/policy-language/) policies.

A sample policy may be:

```
package policy

default allow := false

valid_issuer if {
  input.phdr.cwt.iss == "valid_issuer"
}

allow if {
  valid_issuer
}

errors contains "Invalid issuer" if { not valid_issuer }
```

See [documentation](https://github.com/microsoft/scitt-ccf-ledger/blob/a33698eb4ad1b4cdac3091857f1fcf529304c173/docs/configuration.md) for more detail.

All JS policy testcases have been templatised to run a rego equivalent, including performance regression test_perf, which feeds into the [bencher dashboard](https://bencher.dev/perf/scitt-ccf-ledger). Performance tests show that equivalent Rego policies cause no performance regression compared to JS.

This replaces #260.

Compared to #325, the build and supply chain story is simpler. No additional build tooling is necessary, and this does not introduce dependencies on external crates or non-Microsoft repositories.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,7 +2,7 @@ ARG CCF_VERSION="6.0.23"
 
 # Pin a specific Azure Linux base image tag for reproducibility.
 # Dependabot will monitor and update this FROM line.
-FROM mcr.microsoft.com/azurelinux/base/core:3.0.20260107 AS base
+FROM mcr.microsoft.com/azurelinux/base/core:3.0.20260204 AS base
 
 # Initial stage to download and extract the tdnf snapshot time
 FROM base AS snapshot-extractor


app/src/main.cpp
```diff
@@ -286,8 +286,39 @@ namespace scitt
           throw BadRequestCborError(errors::InvalidInput, e.what());
         }
 
-        if (cfg.policy.policy_script.has_value())
+        // Use Rego policy if defined, otherwise use JS policy if defined
+        // If neither is defined, do not apply any policy, but reject if
+        // CWT issuer is present.
+        if (cfg.policy.policy_rego.has_value())
```
bureado: Would it be accurate to say that this `if` cleanly separates the Rego and JavaScript flows, in that a ledger that only has a JavaScript policy will never hit `check_for_policy_violations_rego()` and therefore never enter the Rego functions in `policy_engine.h`? I'm asking to understand if there's enough modularization of the Rego feature.

I _think_ there is (I don't know much about C++ project architecture) and the next level of modularization I can think of would be making Rego a build time configure option.

test/test_perf.py
achamayou:
```suggestion
    policy = TEST_POLICIES[test_name]
    policy_config = (
        {"policyRego": policy} if "rego" in test_name else {"policyScript": policy}
    )

    configure_service({"policy": policy_config})
```

ivarprudnikov: Please also update `test_write_throughput` which was added recently to track relative throughput based on signature and now will also include the policy type.

PR #353, opened 2026-01-30
steps to sign with cosesigntool

ivarprudnikov:


demo/signing-and-policies/README.md
```diff
@@ -0,0 +1,76 @@
+# Preparing signatures for the service
+
+Signature envelopes (payloads) must be COSE_Sign1 structures as defined in [RFC 8152](https://datatracker.ietf.org/doc/html/rfc8152) but with the specific claims as documented in the [inputs explanation](../../docs/inputs.md).
```
achamayou:
```suggestion
Signature envelopes (payloads) must be COSE_Sign1 structures as defined in [RFC 9052](https://datatracker.ietf.org/doc/html/rfc9052) but with the specific claims as documented in the [inputs explanation](../../docs/inputs.md).
```

There are number of Errata in 9052, so it's better to point at it than at 8152.

PR #390, opened 2026-04-11
Update CCF to v7 and align API with SCRAPI v09

### LINKED ISSUE #377: GetEntry requires retry after successful CreateEntry — violates caller expectation of immediate readability
## Expected Behavior

When CreateEntry succeeds (operation status "succeeded"), an immediate GET /entries/{txid}/statement should return the entry and receipt without requiring retry. A successful write should imply the entry is readable.

## Actual Behavior

GET /entries/{txid}/statement returns 503 Service Unavailable with Retry-After: 1 and error code TransactionNotCached, even though the operation has already reported success.

## Sequence of Events

1. Client calls POST /entries with a signed statement.
2. Client receives 202 Accepted and polls /operations/{txid}.
3. The operations endpoint returns "Status": "succeeded" once the transaction is consensus-committed and indexed.
4. Client immediately calls GET /entries/{txid}/statement.
5. The entry endpoint calls state_cache.get_state_at() — this is the first access for that sequence number, so it kicks off background hydration but returns nullptr.
6. The endpoint returns 503 Service Unavailable with Retry-After: 1 and error code TransactionNotCached.
7. The client must retry to eventually get a 200.

The historical state is typically ready within tens of milliseconds of the first get_state_at() call, but the caller never gets that chance — the endpoint fails immediately on the first nullptr.

## Impact

Spec-compliant HTTP clients — including the Azure SDK — interpret Retry-After: 1 as a full 1-second wait per RFC 7231. Combined with standard exponential backoff strategies, the actual delay is **1–3 seconds per transaction**.

In high-throughput signing and transparency workloads, this per-transaction penalty is unacceptable. Every submitted statement must be followed by receipt retrieval, so the 1–3 second dead time dominates end-to-end latency and makes the service unsuitable for performance-sensitive registration pipelines.

The pyscitt Python client works around this with a dedicated get_historical() helper that retries on TransactionNotCached with tight polling, but this workaround is not available to other SDK consumers and papers over a gap that should not exist in the API contract.

## Relevant Code

- **pp/src/historical/historical_queries_adapter.h** — get_historical_entry_state() calls state_cache.get_state_at() exactly once and throws ServiceUnavailableCborError immediately on 
ullptr.
- **pp/src/operations_endpoints.h** — lookup() reports "succeeded" based on consensus commit and indexing status only; historical state cache readiness is not a precondition for success.
- **pyscitt/pyscitt/client.py** — get_historical() retries on (503, "TransactionNotCached") with a tight loop, masking the issue for Python consumers.

### LINKED ISSUE #394: Align API with SCRAPI draft v9 before GA
Currently we support SCRAPI v3/4 as documented in  https://github.com/microsoft/scitt-ccf-ledger/blob/main/docs/scitt.md

Latest SCRAPI version is at v9 https://www.ietf.org/archive/id/draft-ietf-scitt-scrapi-09.txt 

Alignment must not break the existing clients but also make the API compatible with v9.

### 1. Changing key discovery API

The current implementation exposes key discovery through `/jwks` and `/.well-known/transparency-configuration`.

To align with v09, implement these resources instead:

- `GET /.well-known/scitt-keys`
  - Return a COSE Key Set.
  - Use `Content-Type: application/cbor`.
- `GET /.well-known/scitt-keys/{kid_value}`
  - Return a single COSE Key.
  - Use `Content-Type: application/cbor`.
  - Return `404` with concise problem details when the key does not exist.

Additional requirements:

- If `kid` values are not URL-safe, accept base64url encoding without padding in the path.
- Prefer COSE Key Thumbprint-based `kid` values.
- Keep old keys available long enough for previously issued receipts to remain verifiable.

### 2. Changes in registration flow

The current implementation models asynchronous registration as an operation resource under `/operations/{txid}` and returns a CBOR status document.

To align with v09:

- New sync behavior needs to be added which will be used as a default for the new APIs
- Async behavior would need to change but we'll leave it as is for older clients:
  - `POST /entries` must return `303 See Other` when registration is still running.
  - The `Location` header must point to an `/entries/...` locator, not to `/operations/{txid}`.
  - The running response should not return the current CBOR operation object.

#### 2.1. Add new sync flow

SCRAPI v09 keeps `POST /entries` as the registration endpoint and explicitly allows a synchronous success path.

When the service can produce the receipt within a reasonable time, the expected behavior is:

- `POST /entries`
  - Return `201 Created`.
  - Return the receipt body as `application/cose`.
  - Return `Location: /entries/{entry-id}` or another valid receipt URL.

#### 2.2. Polling changes in async flow (leaving as-is for older clients)

The current implementation uses `GET /operations/{txid}` to poll registration status. We do not change it in favor of saying that only sync behavior will be used for newer API.

To align with v09, polling would need to happen on the entry locator returned by `POST /entries`. This means the current `/operations/{txid}` resource is no longer part of the interoperable API.

- `GET /entries/{locator}` while still running:
  - Return `302 Found`.
  - Return `Location` pointing to the current polling target.
  - Optionally return `Retry-After`.
- `GET /entries/{locator}` when registration completes:
  - Return `200 OK`.
  - Return the receipt as `application/cose`.
  - Optionally return `Location` for the final stable receipt URL.

#### 2.3. Remove the operation-status document from the normative API (leaving as-is for older clients)

The current implementation returns a CBOR object with fields such as `OperationId`, `Status`, `EntryId`, and `Error`.

SCRAPI v09 does not define this response shape. To align with v09:

- Do not use the operation-status CBOR document as part of the standard API.
- If retained, treat it as a non-standard extension only.

### 3. Align error handling on the new resources

The service already uses `application/concise-problem-details+cbor`, which is compatible with v09.

The remaining work is to ensure the new v09 resources return the expected errors for these cases:

- invalid locator
- unknown polling target
- unknown key at `/.well-known/scitt-keys/{kid_value}`
- optional `429 Too Many Requests` during aggressive polling

## Existing endpoints that become non-standard extensions

These endpoints are outside the v09 mandatory API surface and should be documented as extensions if they remain available:

- `/parameters`
- `/parameters/historic`
- `/configuration`
- `/version`
- `/jwks`
- `/.well-known/transparency-configuration`
- `/operations/{txid}`
- `/entries/{txid}/statement`
- `/entries/txIds`

## Practical implementation summary

The main alignment work is:

1. Add `/.well-known/scitt-keys` and `/.well-known/scitt-keys/{kid_value}` using COSE keys.
2. Add synchronous registration to `POST /entries`
3. Reclassify the existing convenience endpoints as implementation-specific extensions.

andpiccione:
- Upgrade CCF from v6 to v7.0.2 (remove enclave/platform abstractions, use ccf::run entrypoint, replace `t_cose` with CCF's built-in COSE verification utilities, etc.)
- Align API with SCRAPI draft-09 (#394) - See #393 and #395
- Add synchronous registration via POST /entries?waitForCommit=true (201 + receipt) - See #392
- Add new API version 2026-03-26 to gate new feature and APIs - See #397 
- Remove /parameters and /parameters/historic endpoints (including related indexing strategies)
- Update pyscitt client, CLI, and test infrastructure for CCF 7 and SCRAPI v09


app/src/service_endpoints.h
```diff
@@ -51,9 +51,30 @@ namespace scitt
       for (const auto& service_certificate : service_certificates)
       {
         auto verifier = ccf::crypto::make_unique_verifier(service_certificate);
+        auto pub_pem = verifier->public_key_pem();
```
achamayou: The new network identity subsystem supports retrieving these keys via get_trusted_keys() now: https://github.com/microsoft/CCF/blob/515e2d6e1fc085c3ae6f294dfec35954359e3ec5/src/node/rpc/network_identity_subsystem.h#L216

It should not be enough to iterate on those and call public_key_jwk(kid) on the public key ptrs it returns. The indexer is no longer necessary.

app/CMakeLists.txt
```diff
@@ -1,10 +1,14 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 3.25)
 
 include(FetchContent)
 
+# Allow FetchContent_Populate() calls (used for t_cose where we need to
```
achamayou: We should be able to drop t_cose in favour of https://github.com/microsoft/CCF/blob/main/include/ccf/crypto/cose_verifier.h, which no longer requires 3P code and uses verified EverCBOR for parsing.

We should discuss what we need to expose of EverCBOR as well to replace QCBOR, but t_cose at least should be removable already.

app/CMakeLists.txt
```diff
@@ -51,16 +89,20 @@ FetchContent_MakeAvailable(regocpp)
...
+# Mark third-party targets as SYSTEM so their headers (and transitive deps)
+# automatically use -isystem, suppressing warnings we don't control.
+set_property(TARGET trieste PROPERTY SYSTEM TRUE)
+set_property(TARGET json PROPERTY SYSTEM TRUE)
+set_property(TARGET yaml PROPERTY SYSTEM TRUE)
+set_property(TARGET t_cose PROPERTY SYSTEM TRUE)
+set_property(TARGET qcbor PROPERTY SYSTEM TRUE)
+
+set(SCITT_TARGET scitt)
+set_target_properties(${SCITT_TARGET} PROPERTIES OUTPUT_NAME cchost)
```
achamayou, line 73: It would be best to avoid calling the binaries cchost for every app to avoid confusion, scitt-ccf, scitt-ccf-ledger or even mst seem preferable.

app/src/service_endpoints.h
andpiccione: We have a separate task to work on using the new CCF API to retrieve the service keys, and one of our developers is already working on it. This will not be included in this PR.

app/CMakeLists.txt
andpiccione: I will look into this and will let you know if there's anything of EverCBOR that can be exposed to remove QCBOR as well.

app/CMakeLists.txt
andpiccione: This was intentional, as changing the binary name would require renaming many other dependent files downstream, which could lead to additional confusion or possibly breaking things. We will continue using `cchost` since it doesn’t cause any problems and serves as a simple convention for backward compatibility.

app/CMakeLists.txt
andpiccione: I modified the code to use the new `cose_verifier` library and dropped the `t_cose` dep: https://github.com/microsoft/scitt-ccf-ledger/pull/390/commits/dc7a652b4d5ac6330edfe80292b8fe7561b10f75

For removing QCBOR, you should probably move the internal EverCBOR API (ccf/_private/crypto/cbor.h) to a public header, so that we can use the utils to encode / decode CBOR (e.g., `ccf::cbor::parse`, `ccf::cbor::serialize`). It would also be nice if you could expose the `parse_x5chain` util from `ccf/_private/crypto/cose_utils.h` for x5chain decoding, but that's less important.

achamayou:
@copilot replace instances of "7.0.0-rc2" with "7.0.0" now that it has been released.

app/CMakeLists.txt
achamayou: This will lead to binaries for tpal and scitt having the same name potentially though, which can have its own consequences.

achamayou (approved):

.pipelines/obpdocker.yml
```diff
@@ -67,7 +64,7 @@ jobs:
         displayName: 'Build and Push to ACR ${{ parameters.AcrName }}'
         retryCountOnTaskFailure: 1 
         inputs:
-          repositoryName: scitt-${{ parameters.Platform }}
+          repositoryName: scitt
```
ivarprudnikov, line 67: We might want to keep it scitt-snp here instead of creating another image repository. This would also allow us to keep history of the image versions which work on snp.

andpiccione: I agree that keeping the image name consistent would be useful for history retention. I'm just concerned the name could be misleading since it references a specific platform, but the image is now generic is no longer limited to that platform only. Since old images are periodically cleaned up in ACR, I feel the image history may get lost anyway at some point.

@achamayou @ryazhang-microsoft Do you have any thoughts on this?

I’m leaning a bit towards changing it for better clarity, but I don’t have a strong preference, so I’m okay with whichever option you prefer.

achamayou: @andpiccione that's a good point, but only if we add support for other x86_64 platforms - which would have to be TDX, not on the roadmap at this point. For ARM CCA, it will have to be a different image anyway, the binaries can't be the same.

Unless I am missing something, docker is quite good at deduplicating layers, so we could have both at quite a low cost: it would look like two containers and two pushes but scitt and scitt-snp would be the same really. One day there might be scitt-tdx that's also the same, and a scitt-cca that's different. I don't have a strong opinion either.

ivarprudnikov (approved):
Ran the demo and all was good, went through most of the changes and it looks OK to me.

.pipelines/obpdocker.yml
ivarprudnikov: There is an annoying issue with multiple tags in onebranch so that to push both we need to rebuild each so we use one version one tag. 
It is fine if you think this generic one makes sense. I'll resolve the conversation.

PR #392, opened 2026-04-20
Add wait_for_commit query parameter to POST /entries

ryazhang-microsoft:
## Summary

Adds a `wait_for_commit` query parameter to the `POST /entries` endpoint. When `wait_for_commit=true`, the HTTP response is deferred until the transaction reaches global consensus, at which point a COSE receipt is returned inline with HTTP 200.

## Changes

- **app/src/main.cpp**: Uses CCF rc2 per-request `set_consensus_committed_function()` API on `RpcContext` to register a committed callback when the query parameter is present. On commit, builds a COSE receipt and returns it. On invalidation, returns an error.
- **pyscitt/pyscitt/client.py**: Adds `submit_signed_statement_blocking()` method that posts to `/entries?wait_for_commit=true` and returns the receipt directly.
- **test/test_blocking.py**: Functional test that validates the blocking behavior, COSE receipt structure, and cryptographic receipt verification.

## Testing

- All 18 C++ unit tests pass
- All 98 functional tests pass (2 skipped, pre-existing)

### DIFF
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -248,8 +248,7 @@ namespace scitt
         const auto parsed_query =
           ccf::http::parse_query(ctx.rpc_ctx->get_request_query());
         const auto wait_for_commit =
-          get_query_value<bool>(parsed_query, "waitForCommit")
-            .value_or(false);
+          get_query_value<bool>(parsed_query, "waitForCommit").value_or(false);
         if (wait_for_commit)
         {
           ctx.rpc_ctx->set_consensus_committed_function(
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -244,8 +244,7 @@ namespace scitt
 
       verifier = std::make_unique<verifier::Verifier>();
 
-      auto register_signed_statement =
-        [this, &context_](EndpointContext& ctx) {
+      auto register_signed_statement = [this, &context_](EndpointContext& ctx) {
         const auto parsed_query =
           ccf::http::parse_query(ctx.rpc_ctx->get_request_query());
         const auto wait_for_commit =
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -257,16 +257,14 @@ namespace scitt
               if (info.status == ccf::FinalTxStatus::Invalid)
               {
                 info.rpc_ctx->set_response_status(
-                  HTTP_STATUS_INTERNAL_SERVER_ERROR);
+                  HTTP_STATUS_SERVICE_UNAVAILABLE);
                 info.rpc_ctx->set_response_header(
                   ccf::http::headers::CONTENT_TYPE,
-                  ccf::http::headervalues::contenttype::CBOR);
-                nlohmann::json error_body;
-                error_body["error"]["code"] = "TransactionInvalid";
-                error_body["error"]["message"] =
-                  "Transaction was dropped by consensus";
-                info.rpc_ctx->set_response_body(
-                  nlohmann::json::to_cbor(error_body));
+                  cbor::CBOR_ERROR_CONTENT_TYPE);
+                info.rpc_ctx->set_response_body(cbor::cbor_error(
+                  errors::InternalError,
+                  "The submission could not be committed and was rolled "
+                  "back, please retry"));
                 return;
               }
 
@@ -279,7 +277,7 @@ namespace scitt
 
               auto cose_receipt = get_cose_receipt(receipt);
 
-              info.rpc_ctx->set_response_status(HTTP_STATUS_OK);
+              info.rpc_ctx->set_response_status(HTTP_STATUS_CREATED);
               info.rpc_ctx->set_response_header(
                 ccf::http::headers::CONTENT_TYPE,
                 ccf::http::headervalues::contenttype::COSE);
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -248,7 +248,7 @@ namespace scitt
         const auto parsed_query =
           ccf::http::parse_query(ctx.rpc_ctx->get_request_query());
         const auto wait_for_commit =
-          get_query_value<bool>(parsed_query, "wait_for_commit")
+          get_query_value<bool>(parsed_query, "waitForCommit")
             .value_or(false);
         if (wait_for_commit)
         {
@@ -439,7 +439,7 @@ namespace scitt
         operation_locally_committed_func,
         authn_policy)
         .add_query_parameter<bool>(
-          "wait_for_commit",
+          "waitForCommit",
           ccf::endpoints::QueryParamPresence::OptionalParameter)
         .set_forwarding_required(ccf::endpoints::ForwardingRequired::Always)
         .set_redirection_strategy(
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -244,7 +244,52 @@ namespace scitt
 
       verifier = std::make_unique<verifier::Verifier>();
 
-      auto register_signed_statement = [this](EndpointContext& ctx) {
+      auto register_signed_statement =
+        [this, &context_](EndpointContext& ctx) {
+        const auto parsed_query =
+          ccf::http::parse_query(ctx.rpc_ctx->get_request_query());
+        const auto wait_for_commit =
+          get_query_value<bool>(parsed_query, "wait_for_commit")
+            .value_or(false);
+        if (wait_for_commit)
+        {
+          ctx.rpc_ctx->set_consensus_committed_function(
+            [&context_](ccf::endpoints::CommittedTxInfo& info) {
+              if (info.status == ccf::FinalTxStatus::Invalid)
+              {
+                info.rpc_ctx->set_response_status(
+                  HTTP_STATUS_INTERNAL_SERVER_ERROR);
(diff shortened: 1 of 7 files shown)


app/src/main.cpp
```diff
@@ -244,7 +244,52 @@ namespace scitt
 
       verifier = std::make_unique<verifier::Verifier>();
 
-      auto register_signed_statement = [this](EndpointContext& ctx) {
+      auto register_signed_statement =
+        [this, &context_](EndpointContext& ctx) {
+        const auto parsed_query =
+          ccf::http::parse_query(ctx.rpc_ctx->get_request_query());
+        const auto wait_for_commit =
+          get_query_value<bool>(parsed_query, "wait_for_commit")
```
ivarprudnikov: Should be blocking by default, no need to have a separate argument, we should follow the proposed API which does not have such an argument in the RFC, see https://www.ietf.org/archive/id/draft-ietf-scitt-scrapi-08.html#section-3.3

app/src/main.cpp
```diff
@@ -244,7 +244,52 @@ namespace scitt
...
+            .value_or(false);
+        if (wait_for_commit)
+        {
+          ctx.rpc_ctx->set_consensus_committed_function(
+            [&context_](ccf::endpoints::CommittedTxInfo& info) {
+              if (info.status == ccf::FinalTxStatus::Invalid)
+              {
+                info.rpc_ctx->set_response_status(
+                  HTTP_STATUS_INTERNAL_SERVER_ERROR);
```
ivarprudnikov: Why is it 500 error? If the user can retry we should return 503 instead.

app/src/main.cpp
```diff
@@ -244,7 +244,52 @@ namespace scitt
...
+                info.rpc_ctx->set_response_header(
+                  ccf::http::headers::CONTENT_TYPE,
+                  ccf::http::headervalues::contenttype::CBOR);
+                nlohmann::json error_body;
+                error_body["error"]["code"] = "TransactionInvalid";
+                error_body["error"]["message"] =
+                  "Transaction was dropped by consensus";
```
ivarprudnikov: Message should be meaningful to a user who does not necessarily understand CCF consensus

app/src/main.cpp
```diff
@@ -244,7 +244,52 @@ namespace scitt
...
+          ctx.rpc_ctx->set_consensus_committed_function(
+            [&context_](ccf::endpoints::CommittedTxInfo& info) {
+              if (info.status == ccf::FinalTxStatus::Invalid)
+              {
+                info.rpc_ctx->set_response_status(
+                  HTTP_STATUS_INTERNAL_SERVER_ERROR);
+                info.rpc_ctx->set_response_header(
+                  ccf::http::headers::CONTENT_TYPE,
+                  ccf::http::headervalues::contenttype::CBOR);
+                nlohmann::json error_body;
```
ivarprudnikov: the error must be in CBOR

app/src/main.cpp
```diff
@@ -244,7 +244,52 @@ namespace scitt
...
+              auto receipt =
+                ccf::endpoints::build_receipt_for_committed_tx(context_, info);
+              if (receipt == nullptr)
+              {
+                return;
+              }
+
+              auto cose_receipt = get_cose_receipt(receipt);
+
+              info.rpc_ctx->set_response_status(HTTP_STATUS_OK);
```
ivarprudnikov: should be 201 in this case https://www.ietf.org/archive/id/draft-ietf-scitt-scrapi-08.html#name-status-201-registration-is-

app/src/main.cpp
ivarprudnikov: current implementation has the converter which acts when doing something like `throw BadRequestCborError` or `throw InternalCborError`, no need to do this sort of a conversion here

ivarprudnikov (changes requested):

app/src/main.cpp
ivarprudnikov: when this path becomes the default one then we can get rid of `operations` endpoints as well as they will not be used anymore

app/src/main.cpp
ryazhang-microsoft: addressed

app/src/main.cpp
```diff
@@ -396,6 +438,9 @@ namespace scitt
         register_signed_statement,
         operation_locally_committed_func,
         authn_policy)
+        .add_query_parameter<bool>(
+          "wait_for_commit",
```
andpiccione: We should use camel case to align with the Azure REST API guidelines: https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#http-query-parameters-and-header-values

app/src/main.cpp
```diff
@@ -244,7 +244,49 @@ namespace scitt
...
+              }
+
+              auto receipt =
+                ccf::endpoints::build_receipt_for_committed_tx(context_, info);
+              if (receipt == nullptr)
+              {
+                return;
+              }
+
+              auto cose_receipt = get_cose_receipt(receipt);
```
andpiccione, line 277: Can you confirm from https://datatracker.ietf.org/doc/draft-ietf-scitt-scrapi/ if we need to return only the COSE receipt or the full transparent statement? I believe we need the latter but I didn't get the chance to check.

pyscitt/pyscitt/client.py
```diff
@@ -603,6 +603,29 @@ def submit_signed_statement_and_wait_for_receipt(
         receipt = self.get_receipt(tx)
         return Submission(operation_id, tx, receipt, False)
 
+    def submit_signed_statement_wait_for_commit(
```
andpiccione, line 606: We should update the `pyscitt` CLI (in particular, the `submit` sub-command, and other related methods) to use the new synchronous flow instead of having to poll for a receipt / transparent statement.

app/src/main.cpp
achamayou: The RFC allows both sync and async flows (very deliberately), and does not forbid passing extra arguments, headers etc. Adding this as a gradual opt-in has the benefit of not breaking clients that use the existing SDK I guess, but that's done just as well with api-version=NEW_DATE I imagine.

app/src/main.cpp
achamayou: Receipts: https://www.ietf.org/archive/id/draft-ietf-scitt-scrapi-08.html#section-3.3.1

achamayou (approved):

app/src/main.cpp
ivarprudnikov: I am just arguing for simplicity, lets remove this branch of async if we are not planning to use it, I do not understand what the value is to keep it. Not to mention it would need to be updated to align with the latest RFC expectations surrounding async.

Another issue is going to GA with this parameter will make it harder for us to remove it. Doing it now will make our API simpler.

For the query parameter, if we are pedantic then the issue is that if we say we are compliant then any compliant client will not pass this parameter and will always default to an async call (it might be OK). But then our client is not strictly compliant (additional parameter presence) and if you take RFCs strictly then additional query parameters not present in the spec could be rejected, which means our client could potentially not work with other compliant SCITT systems purely because they do not understand this parameter.

ivarprudnikov: I am not gonna hold this, it is not a big deal to me.

ivarprudnikov (approved):

app/src/main.cpp
achamayou: Simplicity sounds good, if supporting both for some period is not necessary, that's great!

For clarification, the RFC does not forbid additional parameters, headers etc. What it mandates is that the API must work as expected without, but since both sync and async are allowed, that would be fine.
The sync flow that demands a parameter would not itself be interoperable (since the parameter is not described in the RFC), but the service would still implement the async flow and therefore be compliant. The RFC only says one of them must be implemented, not both.

Anyhow, if we think the transition is manageable without supporting both simultaneously, that's good, it's definitely simpler.

pyscitt/pyscitt/client.py
achamayou: Does the `pyscitt` release flow work at the moment?

app/src/main.cpp
achamayou: Yes, it's definitely the Azure convention, probably best to stick to it even if this is only temporary.

app/src/main.cpp
ryazhang-microsoft: fixed with 503

app/src/main.cpp
ryazhang-microsoft: addressed

app/src/main.cpp
ryazhang-microsoft: Addressed

pyscitt/pyscitt/client.py
ivarprudnikov: @achamayou not yet, it was on me to update it and half of it is done at the moment

ivarprudnikov:
This PR should solve part of #394

pyscitt/pyscitt/client.py
ryazhang-microsoft: Added a flag in pyscitt to allow users to opt in to the new blocking feature. Also updated the documentation.

andpiccione (approved):

PR #393, opened 2026-04-22
User/sbasavapatna/add transparency service keys endpoint

shivasathwikb:
Implement Transparency Service Keys Endpoints As per IETF Document

Document Link : https://datatracker.ietf.org/doc/draft-ietf-scitt-scrapi/

This pull request transitions the SCITT service and client code from using JSON Web Key Sets (JWKS) to using COSE Key Sets encoded in CBOR, aligning with the latest SCITT API draft. It introduces new endpoints for retrieving trusted service keys in COSE format, updates the Python client and verifier to consume and process these endpoints, and removes legacy JWKS logic. The most important changes are:

**Service-side changes (C++):**
* Add  new endpoints: `/.well-known/scitt-keys` (returns all trusted service keys as a COSE_Key_Set in CBOR) and `/.well-known/scitt-keys/{kid_value}` (returns a single key by kid) (`app/src/service_endpoints.h`).
* Implements utility functions to encode EC public keys as COSE_Key structures and assemble COSE_Key_Sets, including mapping CCF curve IDs to COSE curve identifiers and constructing COSE_Key with proper parameters (`app/src/cbor.h`, `app/src/service_endpoints.h`). [[1]](diffhunk://#diff-a4bf62ca187eb5a5c67dadd7ffbed0ddcfd5a2fef81c8e0695f2892386f4b8d7R219-R324) [[2]](diffhunk://#diff-c077ff9defb5ca811889cad67a0fc1f3289b96fcce738d330eebdf90b6dd3ad5L36-R67)

**Client and verification changes (Python):**
* Updates the Python client to fetch and parse the new COSE Key Set endpoints, replacing the old JWKS logic (`pyscitt/pyscitt/client.py`).
* Refactors the verifier to validate the presence of the correct public key in the COSE Key Set, parse COSE_Key structures, and reconstruct EC public keys for verification, removing all JWKS/JWK dependencies (`pyscitt/pyscitt/verify.py`). [[1]](diffhunk://#diff-5242d415d9441f154540a158329de4e27f167016d4cef577990d02108e4da73bL244-R248) [[2]](diffhunk://#diff-5242d415d9441f154540a158329de4e27f167016d4cef577990d02108e4da73bL262-R288) [[3]](diffhunk://#diff-5242d415d9441f154540a158329de4e27f167016d4cef577990d02108e4da73bL359-R371)

**API and configuration changes:**
* Updates the transparency configuration to advertise the new `jwks_uri` location as `/.well-known/scitt-keys` (`app/src/service_endpoints.h`).

**Code cleanup:**
* Removes the old JWKS indexing strategy and related code from the service implementation (`app/src/service_endpoints.h`). [[1]](diffhunk://#diff-c077ff9defb5ca811889cad67a0fc1f3289b96fcce738d330eebdf90b6dd3ad5L36-R67) [[2]](diffhunk://#diff-c077ff9defb5ca811889cad67a0fc1f3289b96fcce738d330eebdf90b6dd3ad5L183-L191)
* Removes unused imports and JWKS-related test code (`test/test_ccf.py`).
* Removes the /parameters endpoint and related code from the same.

**Tests Added**

test_scitt_keys - Verifies endpoint returns correct COSE_Keys
test_scitt_key_by_kid - Verifies single key retrieval
test_scitt_key_by_kid_not_found - Verifies 404 for unknown keys

### DIFF
--- a/app/src/service_endpoints.h
+++ b/app/src/service_endpoints.h
@@ -3,15 +3,15 @@
 
 #pragma once
 
-#include "did/document.h"
+#include "cbor.h"
 #include "visit_each_entry_in_value.h"
 
 #include <ccf/base_endpoint_registry.h>
 #include <ccf/cose_signatures_config_interface.h>
 #include <ccf/crypto/verifier.h>
 #include <ccf/endpoint.h>
-#include <ccf/http_accept.h>
 #include <ccf/json_handler.h>
+#include <ccf/network_identity_interface.h>
 #include <ccf/service/tables/service.h>
 
 namespace scitt
@@ -33,74 +33,39 @@ namespace scitt
   }
 
   /**
-   * An indexing strategy collecting service keys used to sign receipts.
+   * Compute a kid from a public key using SHA-256 hash of the DER encoding.
    */
-  class ServiceKeyIndexingStrategy
-    : public VisitEachEntryInValueTyped<ccf::Service>
+  static std::string kid_from_key(const ccf::crypto::ECPublicKeyPtr& key)
   {
-  public:
-    ServiceKeyIndexingStrategy() :
-      VisitEachEntryInValueTyped(ccf::Tables::SERVICE)
-    {}
+    auto der = key->public_key_der();
+    return ccf::crypto::Sha256Hash(der).hex_str();
+  }
 
-    nlohmann::json get_jwks() const
-    {
-      std::lock_guard guard(lock);
+  /**
+   * Encode an EC public key as a COSE_Key with kid.
+   */
+  static std::vector<uint8_t> key_to_cose_key(
+    const ccf::crypto::ECPublicKeyPtr& key, const std::string& kid)
+  {
+    auto coords = key->coordinates();
+    auto crv = cbor::curve_id_to_cose_crv(key->get_curve_id());
+    return cbor::ec_cose_key_with_kid_to_cbor(crv, coords.x, coords.y, kid);
+  }
 
-      std::vector<nlohmann::json> jwks;
-      for (const auto& service_certificate : service_certificates)
-      {
-        auto verifier = ccf::crypto::make_unique_verifier(service_certificate);
-        auto pub_pem = verifier->public_key_pem();
-        auto kid =
-          ccf::crypto::Sha256Hash(verifier->public_key_der()).hex_str();
-        nlohmann::json json_jwk;
-        try
-        {
-          json_jwk =
-            ccf::crypto::make_ec_public_key(pub_pem)->public_key_jwk(kid);
-        }
-        catch (const std::exception& ec_error)
-        {
-          try
-          {
-            json_jwk =
-              ccf::crypto::make_rsa_public_key(pub_pem)->public_key_jwk(kid);
-          }
-          catch (const std::exception& rsa_error)
-          {
-            throw std::runtime_error(
-              std::string("Unable to parse service public key as EC or RSA. ") +
-              "EC parse failed: " + ec_error.what() +
-              ". RSA parse failed: " + rsa_error.what());
-          }
-        }
-        json_jwk["kid"] = kid;
-        jwks.emplace_back(std::move(json_jwk));
-      }
-      nlohmann::json jwks_json;
-      jwks_json["keys"] = jwks;
-      return jwks_json;
-    }
+  static void set_cbor_response(
+    ccf::endpoints::ReadOnlyEndpointContext& ctx,
+    ccf::http_status status,
+    std::vector<uint8_t>&& body)
+  {
+    ctx.rpc_ctx->set_response_status(status);
+    ctx.rpc_ctx->set_response_header(
+      ccf::http::headers::CONTENT_TYPE,
+      ccf::http::headervalues::contenttype::CBOR);
+    ctx.rpc_ctx->set_response_body(std::move(body));
+  }
 
-  protected:
-    void visit_entry(
-      const ccf::TxID& tx_id, const ccf::ServiceInfo& service_info) override
-    {
-      std::lock_guard guard(lock);
-
-      // It is possible for multiple entries in the ServiceInfo table to contain
-      // the same certificate, eg. if the service status changes. Using an
-      // std::set removes duplicates.
-      service_certificates.insert(service_info.cert);
-    }
-
-  private:
-    mutable std::mutex lock;
-
-    std::set<ccf::crypto::Pem> service_certificates;
-  }; /**
-      * An indexing strategy collecting all past and present service
+  /**
+   * An indexing strategy collecting all past and present service
       * certificates and makes them immediately available.
       */
   class ServiceCertificateIndexingStrategy
@@ -180,15 +145,6 @@ namespace scitt
       out.version = SCITT_VERSION;
       return out;
     };
-
-    static nlohmann::json get_jwks(
-      const std::shared_ptr<ServiceKeyIndexingStrategy>& index,
-      ccf::endpoints::EndpointContext& ctx,
-      nlohmann::json&& params)
-    {
-      // TODO: this is not right when the indexer is not up to date
-      return index->get_jwks();
-    }
   }
 
   static void register_service_endpoints(
@@ -201,12 +157,8 @@ namespace scitt
     auto service_certificate_index =
       std::make_shared<ServiceCertificateIndexingStrategy>();
 
-    auto service_key_index = std::make_shared<ServiceKeyIndexingStrategy>();
-
     context.get_indexing_strategies().install_strategy(
       service_certificate_index);
-
-    context.get_indexing_strategies().install_strategy(service_key_index);
 
     auto get_transparency_config =
       [&](ccf::endpoints::ReadOnlyEndpointContext& ctx) {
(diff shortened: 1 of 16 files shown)


app/src/service_endpoints.h
```diff
@@ -33,76 +33,41 @@ namespace scitt
...
-    std::set<ccf::crypto::Pem> service_certificates;
-  }; /**
-      * An indexing strategy collecting all past and present service
-      * certificates and makes them immediately available.
-      */
+  /**
+   * An indexing strategy collecting all past and present service
+   * certificates and makes them immediately available.
+   */
   class ServiceCertificateIndexingStrategy
```
andpiccione: Do we still need this indexing strategy? I believe this should no longer be needed now that we use the new CCF library.

app/src/service_endpoints.h
```diff
@@ -220,7 +172,8 @@ namespace scitt
 
         nlohmann::json config;
         config["issuer"] = cfg.issuer;
-        config["jwks_uri"] = fmt::format("https://{}/jwks", cfg.issuer);
+        config["jwks_uri"] =
```
andpiccione: Is this the same in the RFC or does it also require update?

andpiccione (approved):

andpiccione:
Tagging @achamayou for review to verify RFC compliance

app/src/service_endpoints.h
shivasathwikb: It is currently used in the **/parameters/historic endpoint,** but I am not certain if we still require this endpoint. We can remove this implementation if the endpoint is no longer in use.

achamayou: It would be preferable for it to be removed, because: 1. it is not necessary to implement SCRAPI and 2. indexing creates a startup cost for nodes linear in the size of the ledger.

app/src/service_endpoints.h
achamayou: The draft does not have a configuration endpoint anymore, only [`scitt-keys`](https://ietf-wg-scitt.github.io/draft-ietf-scitt-scrapi/draft-ietf-scitt-scrapi.html#name-transparency-service-keys) and [scitt-keys/{kid}](https://ietf-wg-scitt.github.io/draft-ietf-scitt-scrapi/draft-ietf-scitt-scrapi.html#name-individual-transparency-ser).

app/src/service_endpoints.h
```diff
@@ -296,25 +249,97 @@ namespace scitt
...
+        cose_keys.push_back(key_to_cose_key(key, kid_from_key(key)));
+      }
+
+      set_cbor_response(
+        ctx, HTTP_STATUS_OK, cbor::cose_key_set_to_cbor(cose_keys));
+    };
+
     registry
-      .make_endpoint(
-        "/jwks",
```
ivarprudnikov, line 306: IMO we cannot just remove it because it is being used by various clients that we do not control at the moment. So the new endpoint just needs adding but not replacing. If we replace this then after the update to the ledger some clients will start failing to verify the receipts.

ivarprudnikov (changes requested):

ivarprudnikov:
This PR should solve part of #394

app/src/service_endpoints.h
```diff
@@ -33,76 +33,41 @@ namespace scitt
...
-      jwks_json["keys"] = jwks;
-      return jwks_json;
-    }
-
-  protected:
-    void visit_entry(
-      const ccf::TxID& tx_id, const ccf::ServiceInfo& service_info) override
-    {
-      std::lock_guard guard(lock);
+    auto der = key->public_key_der();
```
achamayou, line 31: There's a kid from key in CCF, but it doesn't save you much: https://github.com/microsoft/CCF/blob/e6e41e4bfe0b734fff0448340d3947f497cdbe87/src/crypto/public_key.cpp#L10

Still good to switch in case we change it in the future. We can make it accept a `const ECPublicKeyPtr&` too for convenience.

app/src/service_endpoints.h
achamayou: I am confused about how this is going to work, because that endpoint returns COSE Keys now. So surely either the client is new and does not care about the config, or it does, and it will expect jwks at this path, in which case we need to preserve `/jwks` as is, effectively? Am I missing something?

app/src/service_endpoints.h
shivasathwikb: This is still required because it is used by the /jwks endpoint.

app/src/service_endpoints.h
shivasathwikb: I have reverted the changes and added new endpoints: /.well-known/scitt-keys and /.well-known/scitt-keys/{kid_value}.

app/src/service_endpoints.h
shivasathwikb: Keeping this endpoint for backward compatibility as mentioned by Ivar

app/src/service_endpoints.h
shivasathwikb: I added a TODO comment to use the CCF exposed endpoint going forward if it becomes available as the public API, so we can replace the current implementation.

app/src/service_endpoints.h
```diff
@@ -99,10 +101,49 @@ namespace scitt
     mutable std::mutex lock;
 
     std::set<ccf::crypto::Pem> service_certificates;
-  }; /**
-      * An indexing strategy collecting all past and present service
-      * certificates and makes them immediately available.
-      */
```
andpiccione, line 105: I suppose we would need to keep this comment since we are not removing the indexing strategy yet

andpiccione (approved):

app/src/service_endpoints.h
andpiccione: Just to clarify, is it possible to use the new CCF library to get the previous service certificates and return it in JSON format in the `/jwks` endpoint? If so, this would allow us to remove the indexing strategy while keeping the old endpoint. If not, we can leave this as a TODO and revisit it once all clients have moved to the newer SDK version.

shivasathwikb: Removed

app/src/service_endpoints.h
shivasathwikb: ServiceKeyIndexingStrategy and ServiceCertificateIndexingStrategy have both been removed.

app/src/service_endpoints.h
```diff
@@ -342,13 +179,44 @@ namespace scitt
...
+      auto network_identity =
+        context.get_subsystem<ccf::NetworkIdentitySubsystemInterface>();
+      if (!network_identity)
+      {
+        ctx.rpc_ctx->set_response_status(HTTP_STATUS_INTERNAL_SERVER_ERROR);
+        ctx.rpc_ctx->set_response_header(
+          ccf::http::headers::CONTENT_TYPE,
+          ccf::http::headervalues::contenttype::JSON);
+        nlohmann::json error;
+        error["error"] = "Network identity subsystem not available";
```
andpiccione: I would return a more user-friendly error message here

demo/transparency-service-poc/3-client-demo.sh
```diff
@@ -29,9 +29,8 @@ echo -e "\nGetting service parameters"
 SERVICE_PARAMS_FOLDER="$OUTPUT_FOLDER"/service_params
 mkdir -p "$SERVICE_PARAMS_FOLDER"
 
-# Get historic parameters to populate the trust store with all the previous service identities
-# This is useful to verify entries that have been committed with a previous service identity (e.g., prior to a CCF disaster recovery)
-curl -k -f "$SCITT_URL"/parameters/historic > "$SERVICE_PARAMS_FOLDER"/scitt.json
+# Get service parameters to populate the trust store
+curl -k -f "$SCITT_URL"/parameters > "$SERVICE_PARAMS_FOLDER"/scitt.json
```
andpiccione: I think we can get rid of the `/parameters` endpoint as well since it's not used by clients and it just returns information that is already available through the CCF `/node/network` endpoint.

andpiccione: After removing the old endpoints, let's make sure the client demo scripts did not break and everything works well.

pyscitt/pyscitt/verify.py
```diff
@@ -186,7 +186,7 @@ def _parse_service_param(param: dict):
 
                 store[service_id] = ServiceParameters.from_dict(param)
 
-            # If the JSON file contains a list of parameters, parse each one (as returned by /parameters/historic)
+            # If the JSON file contains a list of parameters, parse each one
```
andpiccione: If we remove the `/parameters` endpoint as well, we may need to update pyscitt to no longer use a static trust store for receipt verification.

app/src/service_endpoints.h
achamayou: This should never happen in principle, the subsystem creation is synchronous, the API returns a shared_ptr (as opposed to a reference) for ownership/creation sequence reasons.

pyscitt/pyscitt/verify.py
achamayou: It should be possible to populate the static trust store from `.well-known/scitt-keys` now.

app/src/service_endpoints.h
shivasathwikb: Updated the error message

demo/transparency-service-poc/3-client-demo.sh
shivasathwikb: removed /parameters endpoint

pyscitt/pyscitt/verify.py
shivasathwikb: removed /parameters and replace that by /.well-known/scitt-keys

andpiccione (approved):

andpiccione (approved):

pyscitt/pyscitt/client.py
ivarprudnikov, line 1: pyscitt version needs to be bumped, no? the latest published is at https://pypi.org/project/pyscitt/#history

shivasathwikb: The current version in use is 0.14.0, which is already ahead.

PR #395, opened 2026-04-27
Align registration flow with SCRAPI v09 (sections 2.3-2.5)

ryazhang-microsoft:
## Summary

Aligns the SCITT CCF ledger registration flow with [SCRAPI v09](https://datatracker.ietf.org/doc/draft-ietf-scitt-scrapi/) sections 2.3–2.5, while maintaining **full backward compatibility** with existing clients (including the .NET SDK).

### 1. `POST /entries` — Synchronous flow (201 Created)

| Before | After | Spec (2.3.1) |
|--------|-------|--------------|
| 201 + COSE receipt | 201 + COSE receipt + **`Location: /entries/{txid}`** header | ✅ Aligned |

### 2. `POST /entries` — Asynchronous flow (content negotiation)

The server uses the `Accept` header to determine whether the client expects the legacy or SCRAPI v09 response:

| Client sends | Response | Location | Body | Spec |
|---|---|---|---|---|
| `Accept: application/cbor` (legacy) | **202 Accepted** | `/operations/{txid}` | CBOR `{OperationId, Status}` | Backward compat |
| Anything else / no Accept | **303 See Other** | `/entries/{txid}` | Empty | ✅ SCRAPI v09 §2.3.2 |

### 3. `GET /entries/{txid}` — Polling / Receipt resolution

| Before | After | Spec (2.4 + 2.5) |
|--------|-------|-------------------|
| 503 if tx pending | **302 Found** + `Location` + `Retry-After: 1` | ✅ Aligned |
| 200 + COSE receipt | 200 + COSE receipt (unchanged) | ✅ Already aligned |

### 4. IANA-registered SCITT content types

| Endpoint | Before | After |
|----------|--------|-------|
| `POST /entries?waitForCommit=true` | `application/cose` | `application/scitt-receipt+cose` |
| `GET /entries/{txid}` | `application/cose` | `application/scitt-receipt+cose` |
| `GET /entries/{txid}/statement` | `application/cose` | `application/scitt-statement+cose` |

### Content negotiation on `POST /entries`

Legacy clients (e.g. .NET SDK) that send `Accept: application/cbor` receive the **old 202 + CBOR response** with `Location: /operations/{txid}`. New SCRAPI v09 clients (no Accept or non-CBOR Accept) receive 303 + Location: `/entries/{txid}`.

### `GET /operations/{txid}` kept intact

The legacy `/operations/{txid}` endpoint is **unchanged**. Old clients that poll operation status directly continue to work.

### Python SDK auto-follows redirects

The `request()` method now handles 302/303 redirects natively:
- **302 Found** with `Retry-After`: auto-retried with delay (polling pattern)
- **303 See Other**: auto-followed as GET, dropping the request body
- **307/308**: auto-followed preserving the HTTP method and body

Submit methods use `follow_redirects=False` to inspect the raw 303 and extract the transaction ID.

## Flow Diagram

```
Client                          Server (new)
  |                                |
  |-- POST /entries -------------->|
  |                                |
  |  [waitForCommit=true]          |
  |<--- 201 Created + receipt -----|  (sync, + Location header)
  |   Content-Type: application/scitt-receipt+cose
  |                                |
  |  [default, no Accept: cbor]    |
  |<--- 303 See Other ------------|  (SCRAPI v09)
  |    Location: /entries/2.14     |
  |                                |
  |  [Accept: application/cbor]    |
  |<--- 202 Accepted + CBOR ------|  (legacy compat)
  |    Location: /operations/2.14  |
  |                                |
  |-- GET /entries/2.14 ---------->|
  |<--- 302 Found + Retry-After --|  (pending)
  |                                |
  |-- GET /entries/2.14 ---------->|
  |<--- 200 OK + receipt ----------|  (committed)
  |   Content-Type: application/scitt-receipt+cose
  |                                |
  |-- GET /entries/2.14/statement ->|
  |<--- 200 OK + statement --------|
  |   Content-Type: application/scitt-statement+cose
  |                                |
  |  [.NET SDK / legacy client]    |
  |-- GET /operations/2.14 ------->|  (still works, unchanged)
  |<--- 202 / 200 + CBOR status --|
```

## Files Changed

- `app/src/constants.h` — New `CT_SCITT_RECEIPT` and `CT_SCITT_STATEMENT` constants
- `app/src/main.cpp` — Location header on 201 sync response; SCITT content types on all receipt/statement endpoints
- `app/src/operations_endpoints.h` — Content negotiation: 303 (SCRAPI v09) vs 202 (legacy `Accept: application/cbor`)
- `app/src/historical/historical_queries_adapter.h` — `entry_adapter_with_polling` (302/200)
- `pyscitt/pyscitt/client.py` — Auto-follow 302/303 redirects in `request()`, dual-mode submit, `wait_for_entry()`, SCITT content type constants
- `test/test_cli.py`, `test/test_blocking.py`, `test/test_perf.py` — Tests
- `test/test_scrapi_registration.py` — SCRAPI v09 registration flow + content type tests

## Testing

- 16/16 C++ unit tests pass (2 pre-existing fixture failures)
- 8/8 SCRAPI registration tests pass (including 3 new content type tests)
- 107+ functional tests pass across the full suite

### DIFF
--- a/app/src/historical/historical_queries_adapter.h
+++ b/app/src/historical/historical_queries_adapter.h
@@ -102,4 +102,43 @@ namespace scitt::historical
       f(ctx, state);
     };
   }
+
+  /**
+   * A variant of entry_adapter that returns 302 Found instead of 503
+   * when the transaction is pending or not yet cached, per SCRAPI v09
+   * section 2.4.1. This is used for the GET /entries/{txid} endpoint
+   * to support polling-based registration status queries.
+   *
+   * When the transaction is ready, the handler is called as normal and
+   * returns 200 OK with the receipt (section 2.4.2 / 2.5.1).
+   */
+  static EndpointFunction entry_adapter_with_polling(
+    const HandleHistoricalQuery& f,
+    AbstractStateCache& state_cache,
+    const CheckHistoricalTxStatus& available)
+  {
+    return [f, &state_cache, available](EndpointContext& ctx) {
+      try
+      {
+        auto state = get_historical_entry_state(state_cache, available, ctx);
+        f(ctx, state);
+      }
+      catch (const ServiceUnavailableCborError&)
+      {
+        // Transaction is pending or not yet cached. Per SCRAPI v09
+        // section 2.4.1, return 302 Found with Location pointing to
+        // the same URL, so the client can retry.
+        auto tx_id_str = ctx.rpc_ctx->get_request_path_params().at("txid");
+        ctx.rpc_ctx->set_response_status(HTTP_STATUS_FOUND);
+        if (
+          auto host = ctx.rpc_ctx->get_request_header(ccf::http::headers::HOST))
+        {
+          ctx.rpc_ctx->set_response_header(
+            ccf::http::headers::LOCATION,
+            fmt::format("https://{}/entries/{}", host.value(), tx_id_str));
+        }
+        ctx.rpc_ctx->set_response_header(ccf::http::headers::RETRY_AFTER, "1");
+      }
+    };
+  }
 }
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -280,8 +280,7 @@ namespace scitt
 
               info.rpc_ctx->set_response_status(HTTP_STATUS_CREATED);
               info.rpc_ctx->set_response_header(
-                ccf::http::headers::CONTENT_TYPE,
-                ccf::http::headervalues::contenttype::COSE);
+                ccf::http::headers::CONTENT_TYPE, CT_SCITT_RECEIPT);
               info.rpc_ctx->set_response_header(
                 ccf::http::headers::CCF_TX_ID, info.tx_id.to_str());
               if (host.has_value())
@@ -486,8 +485,7 @@ namespace scitt
 
           ctx.rpc_ctx->set_response_body(cose_receipt);
           ctx.rpc_ctx->set_response_header(
-            ccf::http::headers::CONTENT_TYPE,
-            ccf::http::headervalues::contenttype::COSE);
+            ccf::http::headers::CONTENT_TYPE, CT_SCITT_RECEIPT);
         };
 
       /**
@@ -545,8 +543,7 @@ namespace scitt
 
           ctx.rpc_ctx->set_response_body(statement);
           ctx.rpc_ctx->set_response_header(
-            ccf::http::headers::CONTENT_TYPE,
-            ccf::http::headervalues::contenttype::COSE);
+            ccf::http::headers::CONTENT_TYPE, CT_SCITT_STATEMENT);
         };
 
       /**
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -253,6 +253,8 @@ namespace scitt
         {
           ctx.rpc_ctx->set_consensus_committed_function(
             [&context_](ccf::endpoints::CommittedTxInfo& info) {
+              auto host =
+                info.rpc_ctx->get_request_header(ccf::http::headers::HOST);
               if (info.status == ccf::FinalTxStatus::Invalid)
               {
                 info.rpc_ctx->set_response_status(
@@ -282,6 +284,15 @@ namespace scitt
                 ccf::http::headervalues::contenttype::COSE);
               info.rpc_ctx->set_response_header(
                 ccf::http::headers::CCF_TX_ID, info.tx_id.to_str());
+              if (host.has_value())
+              {
+                info.rpc_ctx->set_response_header(
+                  ccf::http::headers::LOCATION,
+                  fmt::format(
+                    "https://{}/entries/{}",
+                    host.value(),
+                    info.tx_id.to_str()));
+              }
               info.rpc_ctx->set_response_body(cose_receipt);
             });
         }
@@ -480,13 +491,16 @@ namespace scitt
         };
 
       /**
-       * Resolve Receipt, 2.1.4 in
+       * Resolve Receipt / Query Registration Status, 2.4 and 2.5 in
        * https://datatracker.ietf.org/doc/draft-ietf-scitt-scrapi/
+       *
+       * Returns 302 Found with Retry-After if the transaction is still
+       * pending, or 200 OK with the COSE receipt when committed.
        */
       make_endpoint(
         get_entry_receipt_path,
         HTTP_GET,
-        scitt::historical::entry_adapter(
+        scitt::historical::entry_adapter_with_polling(
           get_entry_receipt, state_cache, is_tx_committed),
         authn_policy)
         .set_forwarding_required(ccf::endpoints::ForwardingRequired::Never)
--- a/app/src/operations_endpoints.h
+++ b/app/src/operations_endpoints.h
@@ -479,6 +479,8 @@ namespace scitt
    * In this case, it is assumed the handler has set a response body when it set
    * the erroneous status code.
    *
+   * Per SCRAPI v09 section 2.3.2, async registration returns 303 See Other
+   * with Location pointing to /entries/{txid}.
    */
   static void operation_locally_committed_func(
     ccf::endpoints::CommandEndpointContext& ctx, const ccf::TxID& tx_id)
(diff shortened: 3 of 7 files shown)


ryazhang-microsoft:
@copilot resolve the merge conflicts in this pull request

app/src/operations_endpoints.h
```diff
@@ -479,34 +479,26 @@ namespace scitt
...
-      .status = OperationStatus::Running,
-      .entry_id = {},
-      .error = {}};
-
     ctx.rpc_ctx->set_response_header(ccf::http::headers::CCF_TX_ID, tx_str);
-    ctx.rpc_ctx->set_response_header(
-      ccf::http::headers::CONTENT_TYPE,
-      ccf::http::headervalues::contenttype::CBOR);
-    ctx.rpc_ctx->set_response_body(operation_to_cbor(operation));
-    ctx.rpc_ctx->set_response_status(HTTP_STATUS_ACCEPTED);
```
andpiccione, line 503: Would this be a breaking change for existing clients that expect the operations endpoint here? I suppose not, as long as they can automatically follow redirects

pyscitt/pyscitt/client.py
```diff
@@ -648,6 +706,24 @@ def wait_for_operation(self, operation: str) -> str:
...
+        Returns 302 Found while the registration is still running, and
+        200 OK with the COSE receipt when committed.
+        """
+        resp = self.get(
+            f"/entries/{tx}",
+            retry_on=[
+                HTTPStatus.FOUND.value,
+                HTTPStatus.TOO_MANY_REQUESTS.value,
+                HTTPStatus.SERVICE_UNAVAILABLE.value,
+                (HTTPStatus.SERVICE_UNAVAILABLE, "TransactionNotCached"),
```
andpiccione, line 755: Instead of retrying on the 3xx status codes, wouldn't it be easier to modify the `request` method above to automatically follow redirects instead of retrying?

app/src/operations_endpoints.h
ryazhang-microsoft: Clients that already know the txid can still poll GET /operations/{txid} directly, since that endpoint is unchanged
The breaking change is the response code switch from 202 → 303 and the body going from CBOR → empty. Clients that auto-follow redirects will seamlessly land on GET /entries/{txid}. However, Clients parsing the CBOR body will break unless updated. Let me make the response backward-compatible by using content negotiation. If the client sends Accept: application/cbor (as legacy .NET SDK clients do), return the old 202 + CBOR response. Otherwise, return the new 303.

pyscitt/pyscitt/client.py
ryazhang-microsoft: good point will address

achamayou:
@ryazhang-microsoft there is a registered content type for receipts now: [application/scitt-receipt+cose](https://www.iana.org/assignments/media-types/application/scitt-receipt+cose), which we could and arguably should set in `Content-Type` on what we return, as well as one for statements: [application/scitt-statement+cose](https://www.iana.org/assignments/media-types/application/scitt-statement+cose) which we should accept in `Accept` and arguable encourage as well.

pyscitt/pyscitt/client.py
```diff
@@ -394,18 +397,42 @@ def _get_data():
         while True:
             response = self.session.request(method, url, **kwargs)
 
-            # Handle 307/308 redirects while preserving headers and request metadata
-            if follow_redirects and response.status_code in (307, 308):
+            # Handle redirects (302 Found, 303 See Other, 307/308)
```
achamayou, line 404: It looks like requests has [allow_redirects=True](https://docs.python-requests.org/en/latest/api/#requests.Session.request) as default, and has a [max_redirects](https://docs.python-requests.org/en/latest/api/#requests.Session.max_redirects).

httpx is different and requires passing [follow_redirects=True](https://www.python-httpx.org/compatibility/#redirects). [max_redirects](https://github.com/encode/httpx/blob/master/httpx/_client.py#L198C9-L198C22) can be set when constructing a client.

achamayou (approved):

ryazhang-microsoft:
> @ryazhang-microsoft there is a registered content type for receipts now: [application/scitt-receipt+cose](https://www.iana.org/assignments/media-types/application/scitt-receipt+cose), which we could and arguably should set in `Content-Type` on what we return, as well as one for statements: [application/scitt-statement+cose](https://www.iana.org/assignments/media-types/application/scitt-statement+cose) which we should accept in `Accept` and arguable encourage as well.

Good point! made the change

pyscitt/pyscitt/client.py
ryazhang-microsoft: addressed

app/src/operations_endpoints.h
```diff
@@ -479,34 +479,62 @@ namespace scitt
...
 
-    if (auto host = ctx.rpc_ctx->get_request_header(ccf::http::headers::HOST))
+    // Check if the client is a legacy client expecting CBOR (eg. .NET SDK).
+    // Legacy clients send Accept: application/cbor and expect 202 + CBOR body.
+    auto accept = ctx.rpc_ctx->get_request_header(ccf::http::headers::ACCEPT);
+    bool legacy_client = accept.has_value() &&
```
ivarprudnikov, line 501: This is too brittle, SDK sets multiple accept values, it is better to use API version here
see how .net request is constructed and which parameters,headers are used https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/confidentialledger/Azure.Security.CodeTransparency/src/CodeTransparencyClient.cs#L603-L617

app/src/main.cpp
```diff
@@ -531,8 +543,7 @@ namespace scitt
 
           ctx.rpc_ctx->set_response_body(statement);
           ctx.rpc_ctx->set_response_header(
-            ccf::http::headers::CONTENT_TYPE,
-            ccf::http::headervalues::contenttype::COSE);
+            ccf::http::headers::CONTENT_TYPE, CT_SCITT_STATEMENT);
```
ivarprudnikov, line 546: if cannot just change for all clients and needs to discriminate with the api version to avoid unexpected breaking

app/src/main.cpp
```diff
@@ -475,18 +485,20 @@ namespace scitt
 
           ctx.rpc_ctx->set_response_body(cose_receipt);
           ctx.rpc_ctx->set_response_header(
-            ccf::http::headers::CONTENT_TYPE,
-            ccf::http::headervalues::contenttype::COSE);
+            ccf::http::headers::CONTENT_TYPE, CT_SCITT_RECEIPT);
```
ivarprudnikov, line 488: same thing as below, cannot change for all clients

app/src/main.cpp
```diff
@@ -278,10 +280,18 @@ namespace scitt
 
               info.rpc_ctx->set_response_status(HTTP_STATUS_CREATED);
               info.rpc_ctx->set_response_header(
-                ccf::http::headers::CONTENT_TYPE,
-                ccf::http::headervalues::contenttype::COSE);
+                ccf::http::headers::CONTENT_TYPE, CT_SCITT_RECEIPT);
```
ivarprudnikov, line 283: similar like below comment, we need to discriminate content type with different api versions

app/src/main.cpp
```diff
@@ -475,18 +485,20 @@ namespace scitt
...
-            ccf::http::headervalues::contenttype::COSE);
+            ccf::http::headers::CONTENT_TYPE, CT_SCITT_RECEIPT);
         };
 
       /**
-       * Resolve Receipt, 2.1.4 in
+       * Resolve Receipt / Query Registration Status, 2.4 and 2.5 in
        * https://datatracker.ietf.org/doc/draft-ietf-scitt-scrapi/
+       *
+       * Returns 302 Found with Retry-After if the transaction is still
```
ivarprudnikov, line 495: There is no Retry-After IMO and we should not have it because it forces SDK to actually wait for a minimum of 1sec

pyscitt/pyscitt/client.py
```diff
@@ -332,8 +336,11 @@ def request(
             rolled-back.
 
         follow_redirects:
-            If True (default), automatically follow 307 and 308 redirect responses, preserving
-            all headers and request metadata. If False, return the redirect response as-is.
+            If True (default), automatically follow redirect responses:
+            - 302 Found: follow Location with Retry-After delay (polling pattern)
```
ivarprudnikov, line 340: There is no Retry After and we should not use it to avoid issues in SDK clients

pyscitt/pyscitt/client.py
```diff
@@ -558,13 +592,29 @@ def submit_signed_statement(
...
         headers = {"Content-Type": CT_APPLICATION_COSE}
         resp = self.post(
             "/entries",
             headers=headers,
             content=signed_statement,
+            follow_redirects=False,
         )
+        if resp.status_code == HTTPStatus.SEE_OTHER:
+            location = resp.headers.get("location", "")
+            tx = location.rsplit("/entries/", 1)[-1]
```
ivarprudnikov, line 614: IMO we want to validate this and throw if tx is missing

pyscitt/pyscitt/client.py
```diff
@@ -573,13 +623,31 @@ def submit_signed_statement_and_wait(
...
+            self.wait_for_entry(tx)
```
ivarprudnikov, line 646: IMO need to validate before using

pyscitt/pyscitt/client.py
```diff
@@ -590,13 +658,30 @@ def submit_signed_statement_and_wait_for_receipt(
...
         resp = self.post(
             "/entries",
             headers=headers,
             content=signed_statement,
+            follow_redirects=False,
         )
+        if resp.status_code == HTTPStatus.SEE_OTHER:
+            location = resp.headers.get("location", "")
+            tx = location.rsplit("/entries/", 1)[-1]
+            receipt = self.wait_for_entry(tx)
```
ivarprudnikov, line 681: this bit is repeated, but still needs validation like above

app/src/historical/historical_queries_adapter.h
```diff
@@ -102,4 +102,43 @@ namespace scitt::historical
...
+    AbstractStateCache& state_cache,
+    const CheckHistoricalTxStatus& available)
+  {
+    return [f, &state_cache, available](EndpointContext& ctx) {
+      try
+      {
+        auto state = get_historical_entry_state(state_cache, available, ctx);
+        f(ctx, state);
+      }
+      catch (const ServiceUnavailableCborError&)
```
ivarprudnikov, line 126: must depend on API version as otherwise all clients will start getting HTTP_STATUS_FOUND instead of 503 which is expected by the current clients

PR #397, opened 2026-04-30
Gate SCRAPI v09 behavior on api-version query parameter

ryazhang-microsoft:
## Summary

This PR aligns the SCITT CCF ledger registration flow with **SCRAPI v09 sections 2.3-2.5**, gated behind the `api-version` query parameter for full backward compatibility.

All SCRAPI v9 behavior changes are opt-in via `?api-version=2026-03-26`. Clients that do not send this parameter (or send any other value) continue to get the existing legacy behavior with **zero changes** to request/response format.

## How api-version gating works

| Behavior | `api-version=2026-03-26` (SCRAPI v9) | No api-version / other value (Legacy) |
|---|---|---|
| **POST /entries** (async) | `303 See Other` + `Location: /entries/{txid}` | `202 Accepted` + CBOR body + `Location: /operations/{txid}` |
| **POST /entries?waitForCommit=true** (sync) | `201 Created` with `Content-Type: application/scitt-receipt+cose` | `201 Created` with `Content-Type: application/cose` |
| **Polling (SCRAPI)**: GET /entries/{txid} | `302 Found` + `Location` while pending, `200` with receipt when committed | N/A |
| **Polling (Legacy)**: GET /operations/{txid} | N/A | `503 Service Unavailable` while pending, `200` with CBOR status when committed |
| **GET /entries/{txid}** (committed) | `Content-Type: application/scitt-receipt+cose` | `Content-Type: application/cose` |
| **GET /entries/{txid}/statement** | `Content-Type: application/scitt-statement+cose` | `Content-Type: application/cose` |

### Endpoint flow comparison

**SCRAPI flow** (`api-version=2026-03-26`):
```
POST /entries -> 303 See Other (Location: /entries/{txid})
GET /entries/{txid} -> 302 Found (retry) ... -> 200 OK (receipt)
```

**Legacy flow** (no api-version):
```
POST /entries -> 202 Accepted (Location: /operations/{txid}, CBOR body)
GET /operations/{txid} -> 503 (retry) ... -> 200 OK (CBOR status)
GET /entries/{txid} -> receipt
```

Key distinction: SCRAPI uses `/entries/{txid}` for both polling and receipt retrieval. Legacy uses `/operations/{txid}` for polling and `/entries/{txid}` only for the final receipt.

## Key design decisions

- **api-version over Accept header**: The .NET SDK sends `Accept: "application/cose; application/cbor"` AND `api-version` on every request. Using Accept for gating was brittle and non-deterministic. The api-version query parameter is explicit and already used by the .NET SDK.
- **Unknown versions treated as legacy**: No 400 error for unrecognized api-version values -- they silently fall through to legacy behavior. This avoids breaking clients that send future or unknown versions.
- **No Retry-After header on 302**: Removed per reviewer feedback. The Python client uses a default 2s wait between 302 polls instead.
- **Python client defaults to SCRAPI v9**: `api_version=SCITT_API_VERSION_2026_03_26` by default. Set to `None` for legacy behavior.

### C++ (server-side)
- **`constants.h`**: Added `SCITT_API_VERSION_2026_03_26`, `CT_SCITT_RECEIPT`, `CT_SCITT_STATEMENT` constants
- **`util.h`**: Added `is_scrapi_v9<ContextT>()` template function to check api-version, shared across all endpoints
- **`main.cpp`**: Gated sync 201 content type, GET receipt content type, GET statement content type on `is_scrapi_v9(ctx)`
- **`operations_endpoints.h`**: Gated POST /entries async response on api-version: SCRAPI returns 303 + `Location: /entries/{txid}`, legacy returns 202 + CBOR + `Location: /operations/{txid}`
- **`historical_queries_adapter.h`**: Added `entry_adapter_with_polling()` that returns 302 (SCRAPI) or re-throws 503 (legacy) based on api-version

### Python (client-side)
- **`client.py`**: Added `api_version` parameter to `BaseClient` (defaults to `SCITT_API_VERSION_2026_03_26`); session-level `api-version` query param; tx ID validation with regex; 302 polling with default wait; `_extract_tx_from_location()` helper

### Tests
- **`test_scrapi_registration.py`**: 8 SCRAPI tests using dedicated `scrapi_client` fixture: async 303, 302 polling, sync 201, end-to-end flows, content type assertions
- **`test_cli.py`**: Added `test_submit_default_async_flow` to test the default CLI submit path (async poll without `--wait-for-commit`)
- All **existing tests pass** (110 passed, 2 skipped), confirming backward compatibility
- CI checks pass (clang-format, clang-tidy, black, mypy, isort, copyright)

## Reviewer feedback addressed

- ivarprudnikov: Switch from Accept header to api-version ✅
- ivarprudnikov: Gate 302 vs 503 on api-version ✅
- ivarprudnikov: Remove Retry-After header ✅
- ivarprudnikov: Validate tx ID in Python client ✅
- ivarprudnikov: Rename function to `is_scrapi_v9` ✅
- ivarprudnikov: Default `api_version` to current version ✅
- achamayou: Use httpx follow_redirects ✅
- achamayou: +1 naming to `scrapi_v9` ✅
- andpiccione: IANA-registered SCITT content types ✅
- andpiccione: Rename constant to `SCITT_API_VERSION_2026_03_26` ✅
- andpiccione: Reuse helper function in `operations_endpoints.h` ✅
- andpiccione: Reuse helper function in `historical_queries_adapter.h` ✅

### DIFF
--- a/app/src/constants.h
+++ b/app/src/constants.h
@@ -21,7 +21,7 @@ namespace scitt
   const std::string CT_SCITT_STATEMENT = "application/scitt-statement+cose";
 
   // API versioning for SCITT endpoints.
-  // The new version enables SCRAPI v09 behavior (303, 302, new content types).
+  // SCRAPI v9 enables new behavior (303, 302, new content types).
   // Older versions (or absent api-version) get legacy behavior.
   const std::string SCITT_API_VERSION_2026_03_26 = "2026-03-26";
 
--- a/app/src/constants.h
+++ b/app/src/constants.h
@@ -20,6 +20,11 @@ namespace scitt
   const std::string CT_SCITT_RECEIPT = "application/scitt-receipt+cose";
   const std::string CT_SCITT_STATEMENT = "application/scitt-statement+cose";
 
+  // API versioning for SCITT endpoints.
+  // The new version enables SCRAPI v09 behavior (303, 302, new content types).
+  // Older versions (or absent api-version) get legacy behavior.
+  const std::string SCITT_API_VERSION_SCRAPI = "2026-01-11-preview";
+
   namespace errors
   {
     const std::string IndexingInProgressRetryLater =
--- a/app/src/constants.h
+++ b/app/src/constants.h
@@ -15,6 +15,11 @@ namespace scitt
 
   const std::chrono::seconds OPERATION_EXPIRY{60 * 60};
 
+  // IANA-registered SCITT content types
+  // https://www.iana.org/assignments/media-types/
+  const std::string CT_SCITT_RECEIPT = "application/scitt-receipt+cose";
+  const std::string CT_SCITT_STATEMENT = "application/scitt-statement+cose";
+
   namespace errors
   {
     const std::string IndexingInProgressRetryLater =
--- a/app/src/constants.h
+++ b/app/src/constants.h
@@ -23,7 +23,7 @@ namespace scitt
   // API versioning for SCITT endpoints.
   // The new version enables SCRAPI v09 behavior (303, 302, new content types).
   // Older versions (or absent api-version) get legacy behavior.
-  const std::string SCITT_API_VERSION_SCRAPI = "2026-03-26";
+  const std::string SCITT_API_VERSION_2026_03_26 = "2026-03-26";
 
   namespace errors
   {
--- a/app/src/historical/historical_queries_adapter.h
+++ b/app/src/historical/historical_queries_adapter.h
@@ -102,4 +102,43 @@ namespace scitt::historical
       f(ctx, state);
     };
   }
+
+  /**
+   * A variant of entry_adapter that returns 302 Found instead of 503
+   * when the transaction is pending or not yet cached, per SCRAPI v09
+   * section 2.4.1. This is used for the GET /entries/{txid} endpoint
+   * to support polling-based registration status queries.
+   *
+   * When the transaction is ready, the handler is called as normal and
+   * returns 200 OK with the receipt (section 2.4.2 / 2.5.1).
+   */
+  static EndpointFunction entry_adapter_with_polling(
+    const HandleHistoricalQuery& f,
+    AbstractStateCache& state_cache,
+    const CheckHistoricalTxStatus& available)
+  {
+    return [f, &state_cache, available](EndpointContext& ctx) {
+      try
+      {
+        auto state = get_historical_entry_state(state_cache, available, ctx);
+        f(ctx, state);
+      }
+      catch (const ServiceUnavailableCborError&)
+      {
+        // Transaction is pending or not yet cached. Per SCRAPI v09
+        // section 2.4.1, return 302 Found with Location pointing to
+        // the same URL, so the client can retry.
+        auto tx_id_str = ctx.rpc_ctx->get_request_path_params().at("txid");
+        ctx.rpc_ctx->set_response_status(HTTP_STATUS_FOUND);
+        if (
+          auto host = ctx.rpc_ctx->get_request_header(ccf::http::headers::HOST))
+        {
+          ctx.rpc_ctx->set_response_header(
+            ccf::http::headers::LOCATION,
+            fmt::format("https://{}/entries/{}", host.value(), tx_id_str));
+        }
+        ctx.rpc_ctx->set_response_header(ccf::http::headers::RETRY_AFTER, "1");
+      }
+    };
+  }
 }
--- a/app/src/historical/historical_queries_adapter.h
+++ b/app/src/historical/historical_queries_adapter.h
@@ -2,12 +2,14 @@
 // Licensed under the MIT License.
 #pragma once
 
+#include "constants.h"
 #include "http_error.h"
 #include "tracing.h"
 
 #include <ccf/endpoint_context.h>
 #include <ccf/historical_queries_adapter.h>
 #include <ccf/http_consts.h>
+#include <ccf/http_query.h>
 #include <ccf/json_handler.h>
 #include <ccf/odata_error.h>
 #include <ccf/rpc_context.h>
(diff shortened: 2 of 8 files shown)


app/src/historical/historical_queries_adapter.h
```diff
@@ -125,19 +127,34 @@ namespace scitt::historical
...
-          auto host = ctx.rpc_ctx->get_request_header(ccf::http::headers::HOST))
+        // Transaction is pending or not yet cached.
+        // Check api-version to decide response style.
+        const auto parsed_query =
+          ccf::http::parse_query(ctx.rpc_ctx->get_request_query());
+        auto it = parsed_query.find("api-version");
+        bool scrapi =
+          it != parsed_query.end() && it->second == SCITT_API_VERSION_SCRAPI;
+
+        if (scrapi)
```
andpiccione, line 135: Should we reuse the `is_scrapi_api_version` method here for simplicity?

app/src/operations_endpoints.h
```diff
@@ -495,14 +495,14 @@ namespace scitt
...
-        std::string::npos;
+    // Check if the client requested SCRAPI v09 behavior via api-version.
+    const auto parsed_query =
+      ccf::http::parse_query(ctx.rpc_ctx->get_request_query());
+    auto it = parsed_query.find("api-version");
+    bool scrapi =
+      it != parsed_query.end() && it->second == SCITT_API_VERSION_SCRAPI;
 
-    if (legacy_client)
+    if (!scrapi)
```
andpiccione, line 501: Similar comment as above, here

app/src/constants.h
```diff
@@ -20,6 +20,11 @@ namespace scitt
   const std::string CT_SCITT_RECEIPT = "application/scitt-receipt+cose";
   const std::string CT_SCITT_STATEMENT = "application/scitt-statement+cose";
 
+  // API versioning for SCITT endpoints.
+  // The new version enables SCRAPI v09 behavior (303, 302, new content types).
+  // Older versions (or absent api-version) get legacy behavior.
+  const std::string SCITT_API_VERSION_SCRAPI = "2026-03-26";
```
andpiccione: Instead of labelling this constant with SCRAPI, I'd use a better identifier that explicitly says which Azure API version is being targeted. Something like `SCITT_API_VERSION_2026_03_26` would make more sense, be easier to recognize, and easier to maintain as we add more API versions in the future.

pyscitt/pyscitt/client.py
```diff
@@ -32,6 +33,8 @@
 CBOR_ERR_TITLE_TAG = -1
 CBOR_ERR_DETAIL_TAG = -2
 
+SCITT_API_VERSION_SCRAPI = "2026-03-26"
```
andpiccione: Same comment here

app/src/historical/historical_queries_adapter.h
ryazhang-microsoft: addressed

app/src/constants.h
ryazhang-microsoft: addressed

app/src/util.h
```diff
@@ -3,15 +3,35 @@
...
 {
+  /**
+   * Returns true if the request includes
+   * api-version=SCITT_API_VERSION_2026_03_26.
+   * Used to gate SCRAPI v09 behavior and preserve backward compatibility
+   * for older clients. Unknown or absent api-version values are treated
+   * as legacy.
+   */
+  template <typename ContextT>
+  bool is_scrapi_api_version(const ContextT& ctx)
```
ivarprudnikov: All versions here are SCRAPI, not to mention there is a chance this API will change in the RFC. Calling it like that is misleading a bit, maybe renaming it to scrapi_v9 would make it clearer.

pyscitt/pyscitt/client.py
```diff
@@ -241,6 +245,11 @@ def __init__(
 
         cacert:
             If set and development is False, will be used in TLS verification instead of the default bundle.
+
+        api_version:
+            The api-version query parameter sent with SCITT requests.
+            Set to SCITT_API_VERSION_2026_03_26 for SCRAPI v09 behavior.
+            Defaults to None (legacy behavior).
```
ivarprudnikov: We do not need legacy behavior here, it should default to this current version we are aiming for. Legacy behavior is for currently used SDKs, this cli is not used that much. This means that every test we have will switch to this new API by default which is a good thing IMO

app/src/util.h
achamayou: +1 for scrapi_v9 or draft09 or something similar.

ryazhang-microsoft: Valid point. Address it now.

app/src/util.h
```diff
@@ -3,15 +3,34 @@
...
 namespace scitt
 {
+  /**
+   * Returns true if the request includes the SCRAPI v9 api-version.
+   * Used to gate SCRAPI v9 behavior and preserve backward compatibility
+   * for older clients. Unknown or absent api-version values are treated
+   * as legacy.
+   */
+  template <typename ContextT>
+  bool is_scrapi_v9(const ContextT& ctx)
```
andpiccione, line 26: nit: the API version gating here refers to Azure API versions, not SCRAPI versions. As I mentioned in a previous comment, I recommend framing this method with something that conveys the exact check made (e.g., `is_scitt_api_version_2026_03_26`), since that's the specific check being performed and doesn't have anything to do with SCRAPI imo

app/src/historical/historical_queries_adapter.h
```diff
@@ -129,11 +130,7 @@ namespace scitt::historical
       {
         // Transaction is pending or not yet cached.
         // Check api-version to decide response style.
-        const auto parsed_query =
-          ccf::http::parse_query(ctx.rpc_ctx->get_request_query());
-        auto it = parsed_query.find("api-version");
-        bool scrapi =
-          it != parsed_query.end() && it->second == SCITT_API_VERSION_SCRAPI;
+        bool scrapi = is_scrapi_v9(ctx);
```
andpiccione, line 133: nit: I would use a more precise name for this boolean variable to avoid confusion (since it's not about SCRAPI versions, but rather API versions). Same comment in other parts where a similar variable is used.

andpiccione (approved):

ivarprudnikov (approved):

app/src/operations_endpoints.h
```diff
@@ -483,9 +483,9 @@ namespace scitt
    * with Location pointing to /entries/{txid}.
    *
    * For backward compatibility with legacy clients (eg. .NET SDK) that expect
-   * 202 Accepted with a CBOR body: if the request includes
-   * Accept: application/cbor, the old 202 + CBOR response is returned with
-   * Location: /operations/{txid}.
+   * 202 Accepted with a CBOR body: if the request does not include
+   * api-version=SCITT_API_VERSION_2026_03_26, the old 202 + CBOR response is
```
achamayou:
```suggestion
   * api-version=2026-03-26, the old 202 + CBOR response is
```

achamayou (approved):

app/src/util.h
ryazhang-microsoft: For context, the current name is_scrapi_v9 came from earlier feedback from Ivar and Amaury, who preferred a shorter name that conveys the intent (i.e., "enable the SCRAPI v09 behavior") rather than the exact mechanism. Their reasoning was that is_scrapi_v9 reads more naturally at call sites when gating behavior (e.g., if (is_scrapi_v9(ctx)) { /* SCRAPI flow */ }). I'm happy to rename it if you and the others feel strongly

app/src/historical/historical_queries_adapter.h
ryazhang-microsoft: renamed

Issue #396, opened 2026-04-30
Run e2e regression tests against .net sdk to safeguard existing implementations


ivarprudnikov:
There are some users that rely on azure .net sdk to interact with the ledger. Some API changes might accidentally break the contract which was embedded in that SDK. To make sure we do not introduce compatibility bugs in the future versions download and use .net SDK in the CI tests to check if submission and retrieval of the receipt works as expected. It should be done in a similar spirit to the script `run_functional_tests.sh`

The test script should launch and configure the ledger with the policy to accept the sample payload file. An example test using the payloads we have in the repository is at `test/test_cli.py` named `test_submit_and_validate`.

The implementation will need a small .net console app implementation to run e2e workflow to test these payloads against the running service in the CI server.

SDK source is at https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/confidentialledger/Azure.Security.CodeTransparency 
A simple SDK use example:
```
CodeTransparencyClient client = new(new Uri("https://<< service name >>"));
FileStream fileStream = File.OpenRead("signature.cose");
BinaryData content = BinaryData.FromStream(fileStream);
Operation<BinaryData> operation = await client.CreateEntryAsync(WaitUntil.Started, content);
Response<BinaryData> operationResult = await operation.WaitForCompletionAsync();
string entryId = CborUtils.GetStringValueFromCborMapByKey(operationResult.Value.ToArray(), "EntryId");
Console.WriteLine($"The entry ID to use to retrieve the receipt and transparent statement is {{{entryId}}}");
Response<BinaryData> transparentStatementResponse = client.GetEntryStatement(entryId);
```

PR #398, opened 2026-05-01
add dotnet sdk evaluation regression test

ivarprudnikov:
Related to #396 

- adds a small dotnet program which uses latest preview version of CodeTransparency SDK to submit an entry to the ledger
- adds a python test which uses the dotnet program to submit the example cose file to a configured local ledger
- adds a flag to enable this test on demand
- enables the test in CI environment

### DIFF
--- a/test/e2e_dotnet_sdk/Program.cs
+++ b/test/e2e_dotnet_sdk/Program.cs
@@ -2,7 +2,9 @@
 
 using System.Formats.Cbor;
 using System.Net.Http;
+using System.Net.Security;
 using System.Security.Cryptography.Cose;
+using System.Security.Cryptography.X509Certificates;
 using System.Text;
 using Azure;
 using Azure.Core.Pipeline;
@@ -17,7 +19,7 @@ class Program
     private const string OperationSubmit = "submit";
     private const string OperationVerify = "verify";
     private const string EndpointArgument = "--endpoint";
-    private const string AllowInsecureTlsArgument = "--allow-insecure-tls";
+    private const string CaCertificateArgument = "--ca-certificate";
 
 
     static void Main(string[] args)
@@ -32,13 +34,16 @@ class Program
         }
 
         Uri? endpoint = parsedArguments.Endpoint;
-        bool allowInsecureTls = parsedArguments.AllowInsecureTls;
+        string? caCertificatePath = parsedArguments.CaCertificatePath;
 
         if (endpoint is not null)
         {
             Console.WriteLine($"Using endpoint {{{endpoint}}}");
         }
-        Console.WriteLine($"Development TLS bypass {{{allowInsecureTls}}}");
+        if (caCertificatePath is not null)
+        {
+            Console.WriteLine($"Using CA certificate {{{caCertificatePath}}}");
+        }
 
         Console.WriteLine("Reading the input file...");
         using FileStream fileStream = File.OpenRead(parsedArguments.InputPath);
@@ -55,16 +60,9 @@ class Program
             {
                 options.IdentityClientEndpoint = resolvedIdentityEndpoint.ToString();
             }
-            // In the development environment, we are likely a self-signed certificate
-            if (allowInsecureTls)
+            if (caCertificatePath is not null)
             {
-                HttpClientHandler handler = new()
-                {
-                    ServerCertificateCustomValidationCallback =
-                        HttpClientHandler.DangerousAcceptAnyServerCertificateValidator,
-                };
-                HttpClient httpClient = new(handler);
-                options.Transport = new HttpClientTransport(httpClient);
+                options.Transport = CreateHttpClientTransport(caCertificatePath);
             }
             CodeTransparencyClient client = new CodeTransparencyClient(submitEndpoint, options);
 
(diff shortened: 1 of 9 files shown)


test/test_configuration.py
```diff
@@ -355,6 +358,70 @@ def test_cts_hashv_cwtclaims_payload_with_policy(
...
+        input_receipts = input_message.uhdr.get(crypto.SCITTReceipts, [])
+        output_receipts = output_message.uhdr.get(crypto.SCITTReceipts, [])
+
+        assert output_receipts
+        if input_receipts:
+            assert output_receipts[0] != input_receipts[0]
+
+        output_receipt = pycose.messages.Sign1Message.decode(output_receipts[0])
+        output_cwt = output_receipt.phdr[crypto.CWTClaims]
+        assert output_cwt[crypto.CWT_ISS] == "127.0.0.1:8000"
```
achamayou: Adding a call to the verify operation in dotnet is probably a good idea to exercise the config/jwk endpoints from the SDK perspective as well.

test/e2e_dotnet_sdk/Program.cs
```diff
@@ -0,0 +1,222 @@
...
+            {
+                options.IdentityClientEndpoint = resolvedIdentityEndpoint.ToString();
+            }
+            // In the development environment, we are likely a self-signed certificate
+            if (allowInsecureTls)
+            {
+                HttpClientHandler handler = new()
+                {
+                    ServerCertificateCustomValidationCallback =
+                        HttpClientHandler.DangerousAcceptAnyServerCertificateValidator,
```
achamayou: It would be arguably nicer to pass the server cert instead of disabling verification, but for the purposes of the test this is fine.

test/test_configuration.py
ivarprudnikov: It does not work with the locally issued receipt, I tried that. The issue is that the SDK is verifying the configured sdk client host (`127.0.0.1`) against the issuer in the receipt (`127.0.0.1:8000`) and they never match unless we change the sdk or the issuer in tests (might affect a number of tests).

achamayou (approved):
This is extremely useful to make sure the .NET SDK remains compatible with the service!

test/test_configuration.py
achamayou: Ah the SDK strips the port from the issuer? In production it shouldn't matter, but it'd be good to fix to get the test working. An alternative may be a minimum change to the cchost fixture to manually override the issuer for the service, and to create a per-test fixture just for that test.

test/e2e_dotnet_sdk/Program.cs
ivarprudnikov: now passing the cert instead of ignoring TLS, although TBH all of our python tests ignore TLS at the moment (might be another good improvement for later)

test/test_configuration.py
ivarprudnikov: > the SDK strips the port from the issuer

Kind of, it just compares url.Host to the issuer present in a receipt which has a port in tests. This is not an issue in prod at the moment, but maybe we should be a bit more careful and drop this constraint from the SDK.

I'll try modifying the fixture although it is a bit more complex as the issuer config is part of the config file and not something that we can tweak in config.

achamayou: Yeah it's a bit of a faff, it looks like you need to use the isolated_test mark, and then pass kwargs to that (https://github.com/microsoft/scitt-ccf-ledger/blob/main/test/infra/fixtures.py#L145, https://docs.pytest.org/en/stable/example/markers.html#passing-a-callable-to-custom-markers), then those get fwded to the CCHost constructor from where the config is set. You can then add a set_port_in_issuer=True to CCHost, which controls https://github.com/microsoft/scitt-ccf-ledger/blob/main/test/infra/cchost.py#L388.

ivarprudnikov: This is not gonna be good because when we run functional tests ccf can be started wither via docker (I do that all the time) or through cchost fixture which is used when on azure linux (in CI). Docker has a separate path to configuring the running process which will make it inconsistent if only the fixture is used. I'll better focus on improving the SDK to compare the hosts only without the port and will then be able to come back here to add the verification as well.
SDK has got the verification tests against the mocked expected endpoints already, we do that in pyscitt as well.  If the API contract does not break SDK will work as expected.

achamayou: I'm a bit confused as to why the docker path can't be configured to use the default port if it's on localhost, but anyway, the important thing is that this adds SDK testing!

achamayou (approved):

test/test_configuration.py
ivarprudnikov: Docker is using a template config (i.e. it is static) and there will be no effect for it to be able to use a pytest fixture (docker path starts cchost before tests start). We do use an explicit port already `127.0.0.1:${PORT}` in both cases but what we need for the test is that there should be no port in the issuer i.e. `127.0.0.1`. But IMO even if we set the issuer to `127.0.0.1` the validation will fail because the code will attempt to download the keys using `https://127.0.0.1:443` which will fail as the server is running on another port. Verification test using sdk requires a dedicated separate cchost setup to launch on 443 + setup issuer correctly.

mniamme (approved):

PR #408, opened 2026-06-15
Sync CCF version to 7.0.5, update constitution, and clean up constants

andpiccione:
Update the CCF version to 7.0.5 across relevant files and remove unused constants to streamline the codebase. Additionally, sync the default constitution based on the latest CCF updates.

### DIFF
--- a/app/src/did/document.h
+++ b/app/src/did/document.h
@@ -118,7 +118,9 @@ namespace scitt::did::alt
   struct DIDDocument
   {
     std::string id;
-    std::string context;
+    // @context may be a single string or an array of strings per the W3C DID
+    // spec, so it is stored as raw JSON rather than a fixed type.
+    nlohmann::json context = {};
     std::string type;
     std::vector<DIDDocumentVerificationMethod> verification_method = {};
     nlohmann::json assertion_method = {};
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-ARG CCF_VERSION="7.0.2"
+ARG CCF_VERSION="7.0.5"
 
 FROM mcr.microsoft.com/azurelinux/base/core:3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.18.0]
+### Changes
+- 09b5c99 Update CCF to v7 and align API with SCRAPI v09 (#390)
+- ccd92d2 add dotnet sdk evaluation regression test (#398)
+- 7654281 Bump azurelinux/base/core from 3.0.20260304 to 3.0.20260401 in /docker (#391)
+
 ## [0.17.1]
 ### Changes
 - b7c45ce Update CCF version from 6.0.26 to 6.0.27 (#389)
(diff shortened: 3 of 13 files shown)


app/src/did/document.h
```diff
@@ -118,7 +118,9 @@ namespace scitt::did::alt
   struct DIDDocument
   {
     std::string id;
-    std::string context;
+    // @context may be a single string or an array of strings per the W3C DID
+    // spec, so it is stored as raw JSON rather than a fixed type.
+    nlohmann::json context = {};
```
andpiccione, line 123: This change was necessary due to https://github.com/microsoft/didx509cpp/pull/41 (consumed via CCF 7.0.5). Using `nlohmann::json` should work well, as it handles both strings and arrays. I checked that this should not introduce any backward compatibility issues, since the DIDDocument is only resolved and parsed during the POST request processing and is not stored in the ledger (only the raw COSE statement is persisted). 

@achamayou, could you please confirm that this change is acceptable and will not cause compatibility issues with live instances?

achamayou: I will confirm, but I am fairly sure nothing ever looks at the context, so I doubt this will be a problem.

I think we could and should simplify the code to use resolve_jwk instead of the full document: https://github.com/microsoft/didx509cpp/blob/main/didx509cpp.h#L1735

Constructing the whole document only to extract the key from it again is an unnecessary inefficiency.

achamayou: Yes, it's never read, this is fine.

rjb4standards: Amaury, Could we use context to indicate if the certificate is self-signed or issued from a public CA so that the Transparency service will know how to properly verify the certificate as still valid?

achamayou: Hi @rjb4standards, context here is the JSON-LD context in a DID document (https://www.w3.org/TR/did-1.0/#json-ld), this is a way to define the types used in the DID JSON-LD document, it is not something we can change.

We do not anticipate that Signing Transparency Service will perform freshness checks on certificates for two reasons:

1. There are valid use cases for applying transparency to historical artifacts.
2. The check would provide no values to relying parties, since an arbitrary amount of time may elapse between registration and RP usage.

Relying parties need to check certificate validity when they consume transparent statements if it is necessary to their use cases, this is not something they can delegate to the TS.

achamayou: @andpiccione #409

rjb4standards: Thanks for clarifying Amaury. So the relying party need to perform the CRL check - got it. We can do this, but we will need certificate details indicating how to check for validity(i.e. CRL link). I'm assuming I can get this with a SCRAPI call to get keys, correct?

achamayou: Hi @rjb4standards, the relevant CRL is available from the x5chain that's in the transparent statement protected header (https://www.rfc-editor.org/info/rfc9360/#name-x509-cose-header-parameters).

rjb4standards:
Will I receive the original signing chain for a statement when I call Resolve Receipt for an entry in the SCITT ledger?

achamayou:
> Will I receive the original signing chain for a statement when I call Resolve Receipt for an entry in the SCITT ledger?

@rjb4standards this is a Pull Request to update a dependency of the scitt-ccf-ledger, please use Issues or Discussions for questions.

The Resolve Receipt endpoint is defined here https://ietf-wg-scitt.github.io/draft-ietf-scitt-scrapi/draft-ietf-scitt-scrapi.html#name-resolve-receipt, and the Receipt schema is defined here https://cose-wg.github.io/draft-ietf-cose-merkle-tree-proofs/draft-ietf-cose-merkle-tree-proofs.html#name-usage. The Receipt does not contain duplicate information available in the original Signed Statement.

siddharthUCD (approved):

siddharth9606 (approved):

PR #413, opened 2026-06-30
Cache JS policy interpreter to avoid per-submission recompilation

ivarprudnikov:
Every POST /entries that uses a JavaScript acceptance policy rebuilt a full QuickJS runtime (CommonContext, with all CCF extensions) and recompiled the policy module from source on each submission. The Rego policy path already caches its compiled bundle via a thread_local digest-keyed wrapper; the JS path had no such caching.

Add a thread_local JsPolicyInterpreterCache keyed by the SHA-256 digest of the policy source that retains the interpreter and the compiled `apply` function, rebuilding only when the policy changes. A policy's `apply` function is pure, so reusing the interpreter across submissions is safe; this mirrors the Rego BundleWrapper and the interpreter reuse CCF performs for dynamic JS endpoints. The trade-off is one retained QuickJS runtime per worker thread.

Measured with a new gated microbenchmark (app/unit-tests/policy_engine_perf_test.cpp, 200 iterations, RelWithDebInfo): JS policy evaluation drops from ~91 us to ~1.4 us per call (~65x faster). Also add policy_engine_test.cpp covering accept/reject, repeated invocation, policy-change cache invalidation, and recovery after an invalid policy.

### DIFF
--- a/app/src/policy_engine.h
+++ b/app/src/policy_engine.h
@@ -13,8 +13,11 @@
 #include <ccf/ds/hex.h>
 #include <ccf/js/common_context.h>
 #include <chrono>
+#include <memory>
+#include <optional>
 #include <rego/rego.hh>
 #include <string>
+#include <utility>
 
 namespace
 {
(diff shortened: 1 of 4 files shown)


ivarprudnikov:
Verdict: not safe to merge as-is. The performance goal is valid, but the implementation changes JS policy execution semantics by reusing a mutable QuickJS realm across submissions. That introduces both a regression and a security risk.

- Security angle: A policy could retain one submission’s  payload ,  phdr ,  uhdr , or  details  and expose or use it during a later submission handled by the same worker.
- Recommendation: Do not cache a live mutable JS context/function across submissions. Preserve per-submission JS heap isolation, or cache only immutable compiled artifacts that can be instantiated into a fresh realm.

andpiccione:
> Verdict: not safe to merge as-is. The performance goal is valid, but the implementation changes JS policy execution semantics by reusing a mutable QuickJS realm across submissions. That introduces both a regression and a security risk.
> 
> * Security angle: A policy could retain one submission’s  payload ,  phdr ,  uhdr , or  details  and expose or use it during a later submission handled by the same worker.
> * Recommendation: Do not cache a live mutable JS context/function across submissions. Preserve per-submission JS heap isolation, or cache only immutable compiled artifacts that can be instantiated into a fresh realm.

CCF offers a built-in way to reuse JS interpreters, so maybe we could use that instead of implementing our own logic: https://ccf.dev/main/build_apps/js_app_bundle.html#reusing-interpreters. My understanding is that this would only cache the global state, which shouldn’t be a security concern across different policy executions as long as the input arguments (policy + submission data) are always provided.

@achamayou Would you recommend it?

PR #417, opened 2026-07-14
Update Azure.Security.CodeTransparency SDK to 1.0.0-beta.10 and adapt e2e .NET client

ryazhang-microsoft:
## Summary

Updates the `Azure.Security.CodeTransparency` .NET SDK used by the end-to-end test client (`test/e2e_dotnet_sdk`) from `1.0.0-beta.8` to `1.0.0-beta.10`, adapts the client to the SDK's breaking API changes, and adds regression coverage for both the **synchronous** and **asynchronous** submission paths.

## Why these changes are needed

`1.0.0-beta.10` introduced breaking changes that the e2e client relied on:

1. **`Azure.Security.CodeTransparency.Receipt` namespace removed** — `CborUtils` moved to the root `Azure.Security.CodeTransparency` namespace, so the old `using Azure.Security.CodeTransparency.Receipt;` no longer compiles (`CS0234`).
2. **`CreateEntry(WaitUntil, BinaryData, …)` is now `[Obsolete]`** — the replacement is `CreateEntry(BinaryData, bool waitForCommit, …)` (plus an async `CreateEntryAsync`).
3. **Registration response contract changed** — the new `CreateEntry` no longer returns a CBOR map containing `EntryId`. Its body is now the standalone registration **receipt** (`Content-Type: application/scitt-receipt+cose`), and the entry id (registration transaction id) is obtained from that receipt via `CcfReceipt.GetRegistrationTransactionId(...)`. The previous code parsed `EntryId` out of the `CreateEntry` body and threw at runtime under beta.10 (`the next CBOR data item is of major type '6'`).

## Changes

- **`test/e2e_dotnet_sdk/pipeline-dotnet-cts-cli.csproj`** — bump `Azure.Security.CodeTransparency` to `1.0.0-beta.10`.
- **`test/e2e_dotnet_sdk/Program.cs`**
  - Remove the obsolete `using Azure.Security.CodeTransparency.Receipt;`.
  - Rework the submit flow to the beta.10 contract using only **non-obsolete** APIs: `CreateEntry(content, waitForCommit: true)` → extract the entry id from the receipt with `CcfReceipt.GetRegistrationTransactionId(receipt)` → `GetEntryStatement(entryId)` → write the resulting transparent statement.
  - Add an `--async` flag that routes submission through `CreateEntryAsync` / `GetEntryStatementAsync` (`Main` is now `async Task`).
- **`test/test_configuration.py`** — parametrize `test_dotnet_submit_existing_cose_to_local_service` over `use_async` (`sync`, `async`) so both SDK code paths are exercised by the same assertions.

## Why the SDK flow works (submit → receipt → transparent statement)

- `CreateEntry(content, waitForCommit: true)` registers the signed statement and blocks until it is **committed** to the ledger, returning the **registration receipt** (`application/scitt-receipt+cose`).
- The entry id (registration transaction id) is embedded in the receipt and is read with `CcfReceipt.GetRegistrationTransactionId(receipt)`. This is robust regardless of whether the service commits the entry inline or via a `303 See Other` redirect — the `Location` header is **not** preserved across that redirect, so it cannot be relied upon.
- The standalone receipt is not what downstream verifiers consume. The **transparent statement** is the original `COSE_Sign1` with the receipt embedded in its unprotected header (`SCITTReceipts`). It is produced by the service and downloaded via `GetEntryStatement(entryId)`.
- So the client extracts the entry id from the receipt, calls `GetEntryStatement`, and writes the transparent statement. This preserves the beta.8 end result (a statement carrying an embedded receipt) while using beta.10's new response shape — which is exactly why the existing test assertions (`assert output_receipts`) continue to hold.

## Testing

Validated locally against a real SCITT ledger (built the service image, started it, ran the .NET client through pytest):

- `test_dotnet_submit_existing_cose_to_local_service[sync]` — **PASSED**
- `test_dotnet_submit_existing_cose_to_local_service[async]` — **PASSED**
- `dotnet build` is clean (0 warnings, 0 errors) — no obsolete SDK APIs are used.

### DIFF
--- a/test/e2e_dotnet_sdk/Program.cs
+++ b/test/e2e_dotnet_sdk/Program.cs
@@ -68,17 +68,15 @@ class Program
 
             Console.WriteLine(parsedArguments.UseAsync ? "Sending the signature (async)..." : "Sending the signature...");
             // CreateEntry (with waitForCommit) registers the statement and returns the
-            // registration receipt. The id needed to download the full transparent
-            // statement is returned in the Location response header (".../entries/{entryId}").
-            Response<BinaryData> createEntryResponse = parsedArguments.UseAsync
+            // registration receipt. The entry id (registration transaction id) is
+            // extracted directly from the receipt, which works whether the entry was
+            // committed inline or via a 303 redirect (the Location header is not
+            // preserved across the redirect).
+            Response<BinaryData> receiptResponse = parsedArguments.UseAsync
                 ? await client.CreateEntryAsync(content, true, default)
                 : client.CreateEntry(content, true, default);
-            if (!createEntryResponse.GetRawResponse().Headers.TryGetValue("Location", out string? location)
-                || string.IsNullOrEmpty(location))
-            {
-                throw new Exception("CreateEntry response did not include a Location header with the entry id");
-            }
-            string entryId = location[(location.LastIndexOf('/') + 1)..];
+            BinaryData receipt = receiptResponse.Value;
+            string entryId = CcfReceipt.GetRegistrationTransactionId(receipt.ToArray());
 
             Console.WriteLine($"Waiting for the transparent statement for entry {{{entryId}}}...");
             Response<BinaryData> transparentStatement = parsedArguments.UseAsync
--- a/test/e2e_dotnet_sdk/Program.cs
+++ b/test/e2e_dotnet_sdk/Program.cs
@@ -19,9 +19,10 @@ class Program
     private const string OperationVerify = "verify";
     private const string EndpointArgument = "--endpoint";
     private const string CaCertificateArgument = "--ca-certificate";
+    private const string AsyncArgument = "--async";
 
 
-    static void Main(string[] args)
+    static async Task Main(string[] args)
     {
         ParsedArguments parsedArguments = ParseArguments(args);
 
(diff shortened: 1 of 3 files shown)


andpiccione (approved):

test/e2e_dotnet_sdk/Program.cs
```diff
@@ -66,12 +66,24 @@ static void Main(string[] args)
...
+            // statement is returned in the Location response header (".../entries/{entryId}").
+            Response<BinaryData> createEntryResponse = parsedArguments.UseAsync
+                ? await client.CreateEntryAsync(content, true, default)
+                : client.CreateEntry(content, true, default);
+            if (!createEntryResponse.GetRawResponse().Headers.TryGetValue("Location", out string? location)
+                || string.IsNullOrEmpty(location))
+            {
+                throw new Exception("CreateEntry response did not include a Location header with the entry id");
+            }
+            string entryId = location[(location.LastIndexOf('/') + 1)..];
```
ivarprudnikov: this is incorrect, the location header is lost after the redirect, see sample to check how to obtain the receipt id https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/confidentialledger/Azure.Security.CodeTransparency/samples/Sample1_HelloWorld.md

ivarprudnikov (changes requested):

test/e2e_dotnet_sdk/Program.cs
ryazhang-microsoft: Good catch, thanks! Fixed in 82b02e9 — the entry id is now extracted from the receipt via `CcfReceipt.GetRegistrationTransactionId(receipt)` instead of the `Location` header, so it's correct whether the entry commits inline or via a `303 See Other` redirect (per `Sample1_HelloWorld.md`). Verified locally against a running ledger: both the sync and async submit paths pass.

ivarprudnikov (approved):

PR #420, opened 2026-07-16
Read Endpoints JWT policy enforcement configurable with allowUnauthenticatedReads flag

musabbir:
Adds an optional `allowUnauthenticatedReads` authentication setting for deployments that require JWT on statement registration while allowing unauthenticated access to selected SCITT retrieval endpoints.

## Behavior

`allowUnauthenticatedReads` is optional and inherits `allowUnauthenticated` when omitted, preserving existing behavior:

| `allowUnauthenticated` | `allowUnauthenticatedReads` | `POST /entries` | Selected retrieval endpoints |
| --- | --- | --- | --- |
| `false` | omitted or `false` | JWT required | JWT required |
| `false` | `true` | JWT required | Unauthenticated access allowed |
| `true` | any value | Unauthenticated access allowed | Unauthenticated access allowed |

The selected retrieval endpoints are:

- `GET /entries/{txid}`
- `GET /entries/{txid}/statement`
- `GET /entries/txIds`
- `GET /operations/{txid}`

Service metadata endpoints remain publicly accessible independently of these settings.

## Changes

- Adds the optional configuration field and constitution validation.
- Reuses `ConfigurableEmptyAuthnPolicy` with a read-aware mode.
- Applies the read-aware policy to the selected retrieval endpoints while leaving `POST /entries` on the existing policy.
- Documents the effective configuration modes and public service metadata endpoints.
- Tests every affected retrieval route under inherited, open-read, closed, global-open, and backward-compatible configurations.

## Example: JWT-protected writes with unauthenticated reads

```json
{
  "authentication": {
    "allowUnauthenticated": false,
    "allowUnauthenticatedReads": true,
    "jwt": {
      "requiredClaims": {
        "iss": "https://login.microsoftonline.com/{tenant-id}/v2.0",
        "aud": "https://mst-instance.confidential-ledger.azure.com"
      }
    }
  }
}
```

### DIFF
--- a/app/src/configurable_auth.h
+++ b/app/src/configurable_auth.h
@@ -106,36 +106,20 @@ namespace scitt
    * This policy is always included in the compile time set of policies, but the
    * `authentication.allow_unauthenticated` boolean service configuration option
    * controls whether it is active. The option is false by default.
+   *
+   * When constructed with is_read_policy=true, the policy also checks
+   * `allowUnauthenticatedReads`, enabling a split auth model where write
+   * endpoints require JWT while read endpoints remain publicly accessible.
    */
   class ConfigurableEmptyAuthnPolicy : public ccf::EmptyAuthnPolicy
   {
-    std::unique_ptr<ccf::AuthnIdentity> authenticate(
-      ccf::kv::ReadOnlyTx& tx,
-      const std::shared_ptr<ccf::RpcContext>& ctx,
-      std::string& error_reason) override
-    {
-      auto handle = tx.template ro<ConfigurationTable>(CONFIGURATION_TABLE);
-      auto cfg = handle->get().value_or(Configuration{});
-      if (cfg.authentication.allow_unauthenticated)
-      {
-        return EmptyAuthnPolicy::authenticate(tx, ctx, error_reason);
-      }
-      else
-      {
-        return nullptr;
-      }
-    }
-  };
+    bool is_read_policy_;
 
-  /**
-   * An authentication policy for read endpoints that allows unauthenticated
-   * access based on the `allowUnauthenticatedReads` configuration option.
-   *
-   * This enables a split auth model where write endpoints require JWT
-   * authentication while read endpoints remain publicly accessible.
-   */
-  class ConfigurableEmptyAuthnReadPolicy : public ccf::EmptyAuthnPolicy
-  {
+  public:
+    explicit ConfigurableEmptyAuthnPolicy(bool is_read_policy = false) :
+      is_read_policy_(is_read_policy)
+    {}
+
     std::unique_ptr<ccf::AuthnIdentity> authenticate(
       ccf::kv::ReadOnlyTx& tx,
       const std::shared_ptr<ccf::RpcContext>& ctx,
@@ -143,9 +127,14 @@ namespace scitt
     {
       auto handle = tx.template ro<ConfigurationTable>(CONFIGURATION_TABLE);
       auto cfg = handle->get().value_or(Configuration{});
-      if (
-        cfg.authentication.allow_unauthenticated ||
-        cfg.authentication.allow_unauthenticated_reads)
+
+      bool allow = cfg.authentication.allow_unauthenticated;
+      if (!allow && is_read_policy_)
+      {
+        allow = cfg.authentication.get_allow_unauthenticated_reads();
+      }
+
+      if (allow)
       {
         return EmptyAuthnPolicy::authenticate(tx, ctx, error_reason);
       }
--- a/app/src/configurable_auth.h
+++ b/app/src/configurable_auth.h
@@ -126,4 +126,33 @@ namespace scitt
       }
     }
   };
+
+  /**
+   * An authentication policy for read endpoints that allows unauthenticated
+   * access based on the `allowUnauthenticatedReads` configuration option.
+   *
+   * This enables a split auth model where write endpoints require JWT
+   * authentication while read endpoints remain publicly accessible.
+   */
+  class ConfigurableEmptyAuthnReadPolicy : public ccf::EmptyAuthnPolicy
+  {
+    std::unique_ptr<ccf::AuthnIdentity> authenticate(
+      ccf::kv::ReadOnlyTx& tx,
+      const std::shared_ptr<ccf::RpcContext>& ctx,
+      std::string& error_reason) override
+    {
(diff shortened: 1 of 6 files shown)


app/src/kv_types.h
```diff
@@ -136,6 +136,10 @@ namespace scitt
       JWT jwt;
       bool allow_unauthenticated = false;
 
+      // When true, read endpoints (GET) allow unauthenticated access even
+      // when allow_unauthenticated is false and JWT is configured for writes.
+      bool allow_unauthenticated_reads = true;
```
andpiccione: Defaulting to true is a breaking change with the existing behavior: any service currently running with `allowUnauthenticated: false` will silently start serving reads with no auth, and a brand-new ledger will serve reads openly (which may not be desirable to certain customers, since the current default is not exposing anything to the public unless explicitly configured).

Would it make sense to either:

- default to false and require users to explicitly opt in to open reads
- make the field optional and when unset inherit the value of `allow_unauthenticated` so that we preserve the current behavior?

It's probably not a big deal if no customer is using JWT authentication, but it would be good to confirm this.

app/src/configurable_auth.h
```diff
@@ -126,4 +126,33 @@ namespace scitt
...
+    std::unique_ptr<ccf::AuthnIdentity> authenticate(
+      ccf::kv::ReadOnlyTx& tx,
+      const std::shared_ptr<ccf::RpcContext>& ctx,
+      std::string& error_reason) override
+    {
+      auto handle = tx.template ro<ConfigurationTable>(CONFIGURATION_TABLE);
+      auto cfg = handle->get().value_or(Configuration{});
+      if (
+        cfg.authentication.allow_unauthenticated ||
+        cfg.authentication.allow_unauthenticated_reads)
```
andpiccione: Doesn't this condition mean that setting `allowUnauthenticatedReads: false` will still make read public even though they are explicitly disabled using the new config?

app/src/main.cpp
```diff
@@ -231,6 +231,13 @@ namespace scitt
         std::make_shared<ConfigurableJwtAuthnPolicy>(),
       };
 
+      // Read endpoints use a separate policy that respects
```
andpiccione: I suggest adding some functional tests to validate this new functionality (you're probably already planning to do it)

app/src/kv_types.h
```diff
@@ -177,7 +181,9 @@ namespace scitt
     jwt,
     "jwt",
     allow_unauthenticated,
-    "allowUnauthenticated");
+    "allowUnauthenticated",
+    allow_unauthenticated_reads,
```
andpiccione, line 191: We should update the CCF constitution here and in our internal repository to validate the new field.

app/src/kv_types.h
andpiccione: Let's also update the SCITT configuration document under `docs` to include this new field and explain how it works.

app/src/configurable_auth.h
```diff
@@ -126,4 +126,33 @@ namespace scitt
...
   };
+
+  /**
+   * An authentication policy for read endpoints that allows unauthenticated
+   * access based on the `allowUnauthenticatedReads` configuration option.
+   *
+   * This enables a split auth model where write endpoints require JWT
+   * authentication while read endpoints remain publicly accessible.
+   */
+  class ConfigurableEmptyAuthnReadPolicy : public ccf::EmptyAuthnPolicy
```
andpiccione: This class looks pretty similar to `ConfigurableEmptyAuthnPolicy` with just one extra term in the condition. 

We could consider just using the existing class and simply adding a small flag for the new read config to avoid duplicating the code.

app/src/main.cpp
```diff
@@ -563,7 +570,7 @@ namespace scitt
         HTTP_GET,
         scitt::historical::entry_adapter(
           get_entry_statement, state_cache, is_tx_committed),
-        authn_policy)
+        read_authn_policy)
```
andpiccione, line 573: Should we consider gating the other service endpoints (/version, /jwks, etc.) under the new `read_authn_policy` or are they out of scope? I suppose they are not included on purpose; I'm just flagging this because those are also read endpoints in practice.

app/src/kv_types.h
musabbir: My assumption was no customer is actually using JWT currently. So it works seemlessly in that sense.

app/src/main.cpp
musabbir: I kept them separate as those seem like operation endpoints.

app/src/kv_types.h
musabbir: I switched it to be safe. Likely better coding practice as well.
`allowUnauthenticatedReads` now need to explicity set to true. Otherwise it takes the same value parent.

app/src/kv_types.h
```diff
@@ -177,7 +187,9 @@ namespace scitt
...
+    "allowUnauthenticatedReads");
```
musabbir, line 192: Should this be called `allowPublicReads`

app/src/main.cpp
musabbir: Done

app/src/kv_types.h
rjb4standards: We are not using JWT and have no plans to. 

We are only submitting X.509 identifiers within COSE_Sign1 headers.

We also plan to support ztdnaid URI identifiers that can be verified as trusted PQC compliant identifiers within the Trust Registry as a light weight identity that does not require a Public Key, 
https://www.ietf.org/archive/id/draft-brooks-ztdnaid-new-02.txt

app/src/kv_types.h
musabbir: Keeping `allowUnauthenticatedReads`: it mirrors the existing `allowUnauthenticated` configuration and precisely describes an authentication behavior rather than broader publication semantics. The docs and code comments now define the exact selected SCITT retrieval endpoints it governs.

app/src/kv_types.h
musabbir: Confirmed and completed: the field is optional, inherits `allowUnauthenticated` when omitted, and only opens selected retrieval endpoints when explicitly set to true with `allowUnauthenticated: false`. The PR description, docs, comments, and endpoint-level tests now reflect this contract.

taicchoumsft (approved):

andpiccione (approved):

PR #433, opened 2026-07-29
Clarify and enforce container image reproducibility

ivarprudnikov:
Makes the SCITT container image rebuildable to identical filesystem layers from source, and adds gates in GitHub Actions and Azure DevOps (OneBranch) that fail when that stops being true.

### 1. Core

| File | Why |
| --- | --- |
| `scripts/reproduce-image.sh` (+617) | Single entry point for local, GitHub and OneBranch builds, so the three cannot drift. Builds the context from `git archive` (committed content only) and repacks it with normalized order, ownership, permissions, mtimes and tar header encoding, so `context_sha256` describes the files rather than the build machine. |
| `scripts/compare-image-layers.py` (+87) | Compares `rootfs.diff_ids` between two saved images, Docker or OCI layout. The assertion every gate makes. |

### 2. Making the image bit-identical

| File | Why |
| --- | --- |
| `docker/Dockerfile` (+102) | Base image pinned by digest; `CCF_VERSION` the only CCF input, snapshot time read from that release's `reproduce.json`; `SOURCE_DATE_EPOCH` through `ninja` and `cpack`; final mtime sweep. |
| `docker/canonicalize-image.sh` (+273) | Everything between the last `tdnf` command and the entrypoint install: sorts trust bundles under a certificate-count guard, records the package manifest, rewrites the RPM and tdnf history databases. Rewrites rather than deletes, so the image keeps a queryable package database and working `tdnf history`. |
| `docker/canonicalize-rootfs.sh` (+51) | Normalizes ownership, permissions and mtimes across the rootfs. |
| `app/cpack.cmake` (+7) | Sets RPM `_buildhost` and build time; otherwise every RPM embeds the builder's hostname and clock. |

### 3. Gates

| File | Why |
| --- | --- |
| `.github/workflows/docker-reproducibility.yml` (+351) | Per-commit gate. Builds twice under different path, timezone, locale and umask, compares layers. Publishes `reproduce.json` and `image-layers.txt` on tags. |
| `.github/workflows/docker-reproducibility-historical.yml` (+179) | Weekly rebuild of released tags. The per-commit gate builds twice at one moment, so both builds consume the same changed input and still agree — this is the only gate catching **drift over time**. |
| `.pipelines/obpdocker.yml` (+186) | Same procedure in OneBranch; records verified layer and context digests. |
| `scripts/compare-published-layers.sh` (+137) | Compares built layers against a release's published `image-layers.txt`. Also the documented third-party check. |
| `scripts/compare-github-layers.sh` (+223) | Spans both build systems on every commit. GitHub records its layer digest as a public commit status; OneBranch reads it back, so no credential is needed in a pipeline running YAML from the PR branch. Keyed on the `refs/pull/N/merge` commit, so a moving target branch reports "no record" rather than a false difference. |
| `scripts/check-build-inputs.sh` (+332) | Probes external inputs — base-image tag↔digest consistency, CCF RPM, `reproduce.json` — so a moved input is diagnosed directly instead of as an unexplained layer mismatch. |
| `scripts/check-ccf-version.sh` (+136) | `CCF_VERSION` appears in five files; a devcontainer on a different CCF release reproduces differently for no visible reason. Reads files only, runs first from `ci-checks.sh`. |

### 4. Supporting

- `docker/build.sh` (+22) — `REPRODUCIBLE=1` delegates to the shared script. Stays separate because that path builds only committed content with `--no-cache`.
- `docker/toolchain.env` (+26) — records verified Docker/BuildKit versions; newer warns rather than blocks.
- `docs/reproducibility.md` (+208) — reproduction guide.
- `.pipelines/pullrequest.yml` (−2) — `docker` stage no longer depends on `build`.
- `.gitignore` (+1) — ignores `tmp.*`.
- `.dockerignore` (moved from `docker/`) — Docker only reads it from the context root, so it never took effect before.

## Reviewer notes and known gaps

- Reproducibility depends on upstream availability of the Azure Linux snapshot and CCF RPM. `check-build-inputs.sh` reports this
- Cross-system comparison runs on every PR commit via a public commit status. The production ADO pipeline still needs updating, and will compare against the published `reproduce.json` on tags.
- **Reproduced across machines, not just twice on one runner:** a GitHub run (docker 28.0.4, tar 1.35) rebuilt on an unrelated workstation (docker 29.4.2, tar 1.34) gave the same layers and config digest.
- Tar bytes normalized before hashing, because GNU tar 1.35 writes device fields of ordinary files as empty where 1.34 writes octal zero, so `context_sha256` differed between hosts whose archives extracted to identical trees. 
- The historical workflow can only re-verify tags whose commit already contains `reproduce-image.sh`. Older tags will not be reproducible.
- `CCF_VERSION` is the only CCF input. Possible drift detection moves to recording the hashes, i.e. each build writes the `ccf_reproduce_sha256`, `ccf_rpm_sha256` and `tdnf_snapshottime` it observed, so the historical check still names which input moved. Only the base image stays pinned by digest.


docker/Dockerfile
```diff
@@ -1,27 +1,45 @@
 ARG CCF_VERSION="7.0.10"
-
-# Pin a specific Azure Linux base image tag for reproducibility.
+ARG CCF_RPM_SHA256="4c7f31109aba8893015f4466071d5d964f7117ae0233dd074c96e47eadafe1bc"
+# SHA-256 of the CCF release reproduce.json, and the tdnf snapshot time it
+# contains. The snapshot time determines every package version installed below,
+# so it is pinned here rather than being trusted from an unverified download.
+ARG CCF_REPRODUCE_SHA256="49abb909abeae9abc5481be641513fd3644b5623fdc8769889de95f5489c11c8"
+ARG TDNF_SNAPSHOTTIME="1784205407"
```
andpiccione: Pinning the hardcoded sha256 and snapshot times from the CCF releases adds a bit of friction to CCF version updates and may increase the chance of manual errors.

Is there a reason the previously verified downloads wouldn’t work in this case?

.pipelines/obpdocker.yml
```diff
@@ -73,8 +118,93 @@ jobs:
...
+          fi
+
+          cat "${comparison_output}"
+
+          # Published for cross-checking against the image-layers.txt that the
+          # GitHub release carries for the same commit. The two build systems
+          # use different Docker and BuildKit versions, so this is the only
+          # record that shows whether they still agree.
+          echo "Filesystem layers of the published image:"
+          cat "$(Build.SourcesDirectory)/out/image-layers.txt"
```
andpiccione, line 208: Should we add some check that compares the image layers from GitHub and OneBranch? Could there be a case where the two systems produce different image layers?

docker/Dockerfile
```diff
@@ -51,42 +68,158 @@ RUN mkdir /tmp/app-build && \
...
+        /var/lib/rpm/rpmdb.sqlite-wal && \
+    for state_dir in \
+        /etc/pki/ca-trust/extracted/java \
+        /usr/lib/sysimage/tdnf \
+        /var/cache/ldconfig \
+        /var/cache/tdnf \
+        /var/lib/tdnf; do \
+        if [ -d "${state_dir}" ]; then \
+            find "${state_dir}" -type f \
+                -exec sh -c 'for file do : > "${file}"; done' sh {} +; \
```
andpiccione: While reviewing with Copilot, I noticed that this truncates any file under `tdnf` including the `history.db` file. This causes any subsequent `tdnf` command to exit with error code 9 and prevents operators to install any runtime package on a running container (e.g., for debugging purposes).

.pipelines/obpdocker.yml
ivarprudnikov: I'll explore it a bit more to make sure it is achievable

ivarprudnikov: Done:
1. GitHub Action "Record the verified layers for cross-system comparison" stores the layers in a publicly accessible commit status
2. ADO step `Verify OneBranch image reproducibility`
  - runs 2 docker builds and compares them, then
  - stores layers in a file after running `compare-image-layers.py`, then
  - runs `./scripts/compare-github-layers.sh`, which fetches the GitHub commit status and extracts the layers from it (both GHA and ADO build from the same commit), then compares if the values match

docker/Dockerfile
ivarprudnikov: restored, this type of hardening could be useful but yeah lets not make it too complex, I suppose there are enough controls already to check reproducibility

docker/Dockerfile
ivarprudnikov: This is a difficult one, I kind of needed to wipe them to make sure additional bits do not mess around.

ivarprudnikov: this was reduced and fixed and tdnf works now

scripts/compare-github-layers.sh
```diff
@@ -0,0 +1,223 @@
...
+        echo "The comparison against GitHub Actions was NOT performed. If this" >&2
+        echo "persists, the agent cannot reach the API and this check is not" >&2
+        echo "verifying anything." >&2
+    else
+        echo "GitHub published no ${status_context} record for ${commit} within" >&2
+        echo "${status_timeout}s. Its build may still be running, may have been" >&2
+        echo "superseded, or may never have built this commit." >&2
+        echo "The comparison against GitHub Actions was NOT performed." >&2
+    fi
+    exit 0
```
andpiccione: Should this fail or return a different status code when the comparison was not performed? A missing status, unreachable GitHub API, timeout, or non-success status would all return 0, so OneBranch remains green without comparing its image to the one computed from GitHub.

.github/workflows/docker-reproducibility-historical.yml
```diff
@@ -0,0 +1,179 @@
...
+          echo "Rebuilt manifest:"
+          cat "${rebuilt}"
+
+          if ! gh release download "${ref}" \
+              --repo "${GITHUB_REPOSITORY}" \
+              --pattern reproduce.json \
+              --output "${recorded}" 2> "${RUNNER_TEMP}/download.log"; then
+            cat "${RUNNER_TEMP}/download.log"
+            echo "::notice::${ref} has no published reproduce.json, so only the rebuild itself was verified."
+            exit 0
```
andpiccione, line 211: The job rebuilds before downloading `reproduce.json`, so it re-derives the version and timestamp from today’s Git state instead of using the recorded release inputs. It also exits successfully here if the manifest cannot be downloaded. 

Should we fetch and validate the manifest first, export its recorded inputs, and fail when a release doesn't have the published manifest?

scripts/reproduce-image.sh
```diff
@@ -0,0 +1,617 @@
...
+PY
+}
+
+# The Dockerfile pins the CCF release by version alone, so the checksums that
+# describe what that release actually served are observed here rather than read
+# back out of the build. Recording them is what lets the historical rebuild
+# check say which input moved when an old image stops reproducing, instead of
+# only reporting that the layers differ.
+observed_ccf_reproduce_sha256=""
+observed_ccf_rpm_sha256=""
```
andpiccione, line 441: These hashes are observed after the Docker build through separate downloads. The Dockerfile independently downloads reproduce.json and the RPM during the build, so the resulting manifest is not guaranteed to describe the bytes actually consumed. The RPM value is API metadata rather than a hash of the downloaded build input. 

Should we instead download and hash each input once, then feed those exact files into the build and manifest?

andpiccione (dismissed):

scripts/compare-github-layers.sh
ivarprudnikov: unreachable API exits 1, missing record exits 2, both fail the ADO step; the separate status exists because the same command is documented for third-party use, and only GitHub's own non-success state stays non-fatal

.github/workflows/docker-reproducibility-historical.yml
ivarprudnikov: manifest is fetched before the rebuild and its recorded inputs exported,  source_commit  compared rather than exported; a missing manifest is reported rather than failed, since no release has one yet, but an unreadable one now fails

scripts/reproduce-image.sh
ivarprudnikov: each stage hashes the asset as it downloads it and the runtime stage writes  /opt/scitt/share/build-inputs.json , which the manifest reads back out of the image; hashes match GitHub's published digests exactly and the manifest path makes no network calls

andpiccione (dismissed):

andpiccione (approved):

PR #445, opened 2026-08-25
sync test dependencies to match pyscitt version

ivarprudnikov:


rjb4standards:
Do you know when the new pyscitt will be available for install from PYpi?

Please confirm that there are no changes required to CoseSignTool sign and the new pyscitt will accept the existing output for registration within the Microsoft SCITT Ledger

andpiccione (approved):

ivarprudnikov:
> Do you know when the new pyscitt will be available for install from PYpi?

`pyscitt` was available to be installed from pypi for quite some time, see https://pypi.org/project/pyscitt/ Latest release was yesterday.

> Please confirm that there are no changes required to CoseSignTool sign and the new pyscitt will accept the existing output for registration within the Microsoft SCITT Ledger

`CoseSignTool` is managed separately and has a slightly different set of features to `pyscitt` and there are no controls in place which would guarantee the feature parity between them. But the `submit` operation in `pyscitt` should work with any cose envelope, the validation would otherwise happen on the server.

rjb4standards:
Thanks, Ivar. Does the latest pyscitt product on PyPi implement SCRAPI version 11 semantics and syntax?

PR #449, opened 2026-09-09
Fix uncaught exception when rego policy/errors yields no result

musabbir:
## Summary

Fixes an uncaught exception that surfaces as an opaque HTTP 500
`{code: "InternalError", message: "Index out of range"}` when a rego
registration policy denies a submission (`policy/allow == false`) but its
`policy/errors` rule produces **no result** for that specific input (e.g. a
complete rule guarded by `if` with no matching branch and no `default`).

### Root cause

In `check_for_policy_violations_rego()` (`app/src/policy_engine.h`), after a
well-formed policy denial, the code queries `policy/errors` to build a
human-readable violation message via
`set_from_terms(error_output.expressions())`. When the policy's `errors` rule
is undefined for the given input, `rego-cpp`'s `Output::expressions()` (→
`expressions_at(0)`) throws `std::out_of_range("Index out of range")`.

The surrounding `catch` clause only caught `std::domain_error`, a sibling of
`std::out_of_range` (both derive from `std::logic_error`, not from each
other), so the exception was not caught there. It also isn't caught anywhere
in `main.cpp`, so it escapes all the way to CCF's generic catch-all
(`app/src/http_error.h`), which turns any uncaught `std::exception` into a raw
HTTP 500 `InternalError` using `e.what()` as the message.

This was reported against a real submission whose leaf certificate EKU was
rejected by the tenant's configured policy, and confirmed live in production
logs for the reported `ClientRequestId`:

```
Unhandled exception in endpoint: Code=InternalError Index out of range ClientRequestId=e9287e6c-d0f0-47b9-b646-f1548251d089 RequestId=e09e662911e39f6
```

### Fix

Broaden the `catch` around the `policy/errors` extraction to `std::exception`
(matching the handling already used a few lines above for the `policy/allow`
query), so this class of policy-authoring gap now results in a clean 400
`PolicyError` response instead of an uncaught 500.

### Testing

Added a regression test (`app/unit-tests/policy_engine_test.cpp`) with a rego
policy whose `errors` rule is undefined for the input, asserting a graceful
`BadRequestCborError`/`PolicyError` is thrown rather than letting the
exception escape uncaught.

Opening as a PR (rather than pushing directly to `main`) to trigger CI, since
building/running the C++ unit tests locally isn't practical in this
environment.

### DIFF
--- a/app/src/policy_engine.h
+++ b/app/src/policy_engine.h
@@ -713,8 +713,15 @@ namespace scitt
 
       return buf.str();
     }
-    catch (const std::domain_error& e)
+    catch (const std::exception& e)
     {
+      // error_output.expressions() (i.e. Output::expressions_at(0)) throws
+      // std::out_of_range, not std::domain_error, when the policy/errors
+      // rule produced no result for this input (e.g. the policy only
+      // defines specific violation messages and this denial reason isn't
+      // one of them). Catching std::exception here - rather than only
+      // std::domain_error - ensures any such policy authoring gap is
+      // reported as a 400 PolicyError instead of an uncaught 500.
       throw BadRequestCborError(
         scitt::errors::PolicyError,
         fmt::format("Could not interpret policy errors: {}", e.what()));
--- a/app/src/policy_engine.h
+++ b/app/src/policy_engine.h
@@ -686,8 +686,26 @@ namespace scitt
         fmt::format("Could not interpret policy result: {}", e.what()));
     }
 
-    auto error_output =
-      rego::Output(interpreter.query_bundle(bundle, "policy/errors"));
+    auto error_node = interpreter.query_bundle(bundle, "policy/errors");
+
+    // A policy whose "errors" rule produces no result for this input (e.g.
+    // the policy only defines specific violation messages and this denial
+    // reason isn't one of them) makes run_entrypoint() return the bare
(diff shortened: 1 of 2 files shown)


achamayou (dismissed):
Good catch!

app/src/policy_engine.h
achamayou: @musabbir for the test you have, errors is undefined, and so you would need something like:

```
auto error_node = interpreter.query_bundle(...);

if (error_node != nullptr && error_node->type() == rego::Undefined)
{
  throw BadRequestCborError(...);
}

auto error_output = rego::Output(error_node);

...
```

musabbir: Good catch, thank you! Confirmed the root cause: `run_entrypoint()` returns the bare `Undefined` token node (not wrapped in `Results`) when the entrypoint is undefined for the input, and `rego::Output`'s constructor only expects `Results` -- its internal assert would abort a debug build instead of throwing.

Pushed 6dc2321 implementing exactly what you suggested: check `error_node->type() == rego::Undefined` before constructing the `Output`, and throw `BadRequestCborError(PolicyError, ...)` directly in that case. Kept the existing broad catch around `set_from_terms()` as a secondary safety net for other unexpected shapes (e.g. a defined but non-set value).

achamayou (approved):

PR #452, opened 2026-09-18
[pyscitt] Support raw CBOR protected headers

musabbir:
Custom protected headers can be added to Cose Structure. This change adds the ability to add a raw CBOR in addition to existing functionality.
Useful when adding external signatures as headers.


test/test_encoding.py
```diff
@@ -246,3 +246,37 @@ def test_x5chain(
...
+            content_type="text/plain",
+            additional_phdr={
+                "com.example.signature": crypto.RawCbor(raw_value),
+            },
+        )
+
+        cose_sign1 = cbor2.loads(signed_statement)
+        protected_header = cose_sign1.value[0]
+        assert raw_value in protected_header
+        assert cbor2.loads(protected_header)["com.example.signature"] == {
```
ivarprudnikov, line 275: I think we also want a test which checks the exact bytes instead of a struct, to see what bytes are under `"com.example.signature"` which will allow verifying that it is a bstr which has a map, or if it is a map without bstr. At the moment it is opaque and unclear what the actual cbor data structure gets encoded for this header value.

pyscitt/README.md
```diff
@@ -12,4 +12,22 @@ Package sources are available at https://github.com/microsoft/scitt-ccf-ledger/t
...
+
+```python
+from pyscitt import crypto
+
+signed_statement = crypto.sign_statement(
+    signer,
+    payload,
+    content_type="application/json",
+    additional_phdr={
+        "com.example.signature": crypto.RawCbor(existing_cbor_signature),
```
ivarprudnikov, line 29: I think this is redundant, it is possible to achieve the same with

```
"com.example.signature": cbor2.dumps(existing_cbor_signature),
```

`cbor2` is already our dependency which will exist, not need to create a proxy crypto.RawCbor

Otherwise, nested structures are supported out of the box if bstr encoding is not necessary.

Added more tests to avoid confusion in the future, see https://github.com/microsoft/scitt-ccf-ledger/pull/453/changes

ivarprudnikov (changes requested):

test/test_encoding.py
ivarprudnikov: added more tests in another PR https://github.com/microsoft/scitt-ccf-ledger/pull/453/changes to see what I mean

pyscitt/README.md
musabbir: Well, my goal was to have parity with what is being developed on CoseSignTool so that we dont have to rely on that tool to run tests.

https://github.com/microsoft/cosesigntool/pull/212

ivarprudnikov: `crypto.RawCbor` is just a proxy which uses `cbor2` to encode the data structure into a `bstr`

> so that we dont have to rely on that tool

I am not following, if the user needs to use imports in python shown in the example then they can import cbor2 already, or is this a reference to the CLI command?

I looked at the linked PR but that is like 2k+ changes, not sure what exactly are we trying to match there?

musabbir: CoseSignTool sign `
  --payload app.bin `
  --signature app.cose `
  --pfx cert.pfx `
  --CborProtectedHeaders vendor-signature=RgABAgME   <---- This is what they are adding

 RgABAgME  is base64 for one fully encoded CBOR item. CoseSignTool places that decoded CBOR item directly at the  vendor-signature  protected-header label—without adding a CBOR byte-string wrapper.

I am okay to wait for their PR to merge.

